PR #48, opened 2024-10-27
Add Class metadata models

notaphplover:
### Added
- Added `ClassMetadata` models.
- Added `getClassMetadata`.

### Motivation
In order to migrate `inversify`, several responsibilities are being moved from `inversify` to `@inversifyjs/core`. The first one is metadata management.

Currently, `inversify` has this responsibility distributed among several modules, making hard figuring out what's expected to be stored as metadata:
- It's currently challenging to know what's the expected metadata associated to a type. `inversify` metadata internal data structures don't represent some edge cases (like the handling of lazy service identifiers).
- Several checks as enforcing no duplicated tags can be simplified with the right data structures.
- Code can be decoupled from `reflect-metadata`, making it more maintainable.
- This decoupling would unlock the implementation of better metadata representations. This can result in reducing the operations and the amount of objects involved in the process of planning the instantiation of a service, something that would result in the simplification of the source code and boosting the performance of the library.

- `ClassMetadata` is proposed as a data structure that represents the metadata associated to a certain type. This data structure handles in an efficient way most of the operations involved in the planning process.

### Testability
Once this approach is discussed and validated, several tests must be included:
1. Unit tests should validate all the code flows with a decent mutation testing score.
2. E2E tests should enforce `getClassMetadata` is consistent with `inversify`'s `MetadataReader` and `Target`.


notaphplover:
Hey @rsaz , I've been analyzing the way `inversify` handles metadata in order to delegate that responsibility in this package. I've drafted this proposal, what do you think about it?

packages/container/libraries/core/src/metadata/calculations/getClassElementMetadataFromLegacyMetadata.ts
```diff
@@ -0,0 +1,82 @@
...
+  );
+
+  if (unmanagedMetadata !== undefined) {
+    return {
+      kind: ClassElementMetadataKind.unmanaged,
+    };
+  }
+
+  if (multiInjectMetadata === undefined && injectMetadata === undefined) {
+    throw new Error();
```
notaphplover, line 40: Consider this as sample code, we should throw a meaningful error indicating either `inject` `multiInject` or `unmanaged` must be included

rsaz: Hi @notaphplover 

I read through your proposal and think it makes a lot of sense—it would definitely be a valuable addition to the Inversify core. One point that could enhance your proposal is elaborating a bit more on the project structure. Specifically, how it would look if we were to implement additional packages, such as inversify-bun-utils, inversify-express-utils, or inversify-fastify-utils, with the core package containing the shared decorators for these HTTP frameworks.

Another aspect to consider is a modular approach with:

@inversify/core
@inversify/common

The common module could include utilities that are shared across implementations, such as decorators, without requiring the full core package. This would enable a lightweight, flexible option for those implementing plugins or working within different frameworks, as they could pull from common without the overhead of the entire core.

Overall, the proposal looks strong, and these additional considerations could make it even more versatile. Great work!

PR #59, opened 2024-10-31
feat(strongly-typed): initial commit

alecgibson:
Add an initial offering for `@inversifyjs/strongly-typed`, which will let consumers optionally apply strong typing to their `Container`.

### DIFF
--- a/packages/container/libraries/strongly-typed/README.md
+++ b/packages/container/libraries/strongly-typed/README.md
@@ -1,0 +1,110 @@
+[![Test coverage](https://codecov.io/gh/inversify/monorepo/branch/main/graph/badge.svg?flag=%40inversifyjs%2Fstrongly-typed)](https://codecov.io/gh/inversify/monorepo/branch/main/graph/badge.svg?flag=%40inversifyjs%2Fstrongly-typed)
+[![npm version](https://img.shields.io/github/package-json/v/inversify/monorepo?filename=packages%2Fcontainer%2Flibraries%2Fstrongly-typed%2Fpackage.json&style=plastic)](https://www.npmjs.com/package/@inversifyjs/strongly-typed)
+
+# @inversifyjs/strongly-typed
+
+Type definitions for adding strong typing to the `Container` and `@inject()` decorator.
+
+
+## Getting started
+
+This library can be used in a couple of ways, depending on your preference.
+
+All usages wind up with you having a strongly-typed `TypedContainer`, which is a generic class that accepts a binding map as a type argument, which forms the contract that your container will adhere to:
+
+```ts
+import { TypedContainer } from '@inversifyjs/strongly-typed';
+
+interface Foo { foo: string }
+interface Bar { bar: string }
+
+interface BindingMap {
+  foo: Foo;
+  bar: Bar;
+}
+
+const container = new TypedContainer<BindingMap>();
+
+// Bindings are now strongly typed:
+
+container.bind('foo').toConstantValue({foo: 'abc'}); // ok
+container.rebind('foo').toConstantValue({unknown: 'uh-oh'}) // compilation error
+
+let foo: Foo = container.get('foo') // ok
+foo = container.get('bar') // compilation error
+foo = container.get('unknown-identifier') // compilation error
+```
+
+### Instantiation
+
+The simplest way to use the library is to directly construct a `TypedContainer`:
+
+```ts
+import { TypedContainer } from '@inversifyjs/strongly-typed';
+
+const container = new TypedContainer<BindingMap>();
+```
+
+This class is actually just a re-typed re-export of the vanilla `Container`, so shares all underlying functionality.
+
+### Type assertion
+
+If you'd prefer to keep this library out of your final dependency tree, you can just import the types and perform a type assertion:
+
+```ts
+import { Container } from 'inversify';
+// The type import will be stripped during transpilation
+import type { TypedContainer } from '@inversifyjs/strongly-typed';
+
+const container = new Container() as TypedContainer<BindingMap>;
+```
+
+
+## Advanced usage
+
+### `Promise` bindings
+
+`inversify` allows binding `Promise`s, but these **must** be retrieved using `getAsync()` trying to resolve them using `get()` will throw.
+
+The type signatures will help to enforce this if you show an async binding in the binding map with a `Promise`:
+
+```ts
+interface BindingMap {
+  number: number;
+  asyncNumber: Promise<number>;
+}
+
+const container = new TypedContainer<BindingMap>();
+
+container.get('number') // number
+container.get('asyncNumber') // compiler error
+container.getAsync('asyncNumber') // Promise<number>
+```
+
+### Container hierarchies
+
+A strongly-typed child container can be created from a parent container.
+
+This strong typing is optional for the child:
+
+```ts
+const parent = new TypedContainer<BindingMap>();
+const child = parent.createChild(); // weakly typed
+```
+
+If the child is strongly-typed, its binding will automatically be adjusted to include the parent types:
+
+```ts
+const parent = new TypedContainer<{foo: Foo}>();
+const child = parent.createChild<{bar: Bar}>();
+child.get('bar') // ok
+child.get('foo') // ok
+```
+
+A child may also override a parent's bindings with a completely unrelated type:
+
+```ts
+const parent = new TypedContainer<{foo: Foo}>();
+const child = parent.createChild<{foo: Bar}>();
+const resolved: Bar = child.get('foo');
+```
--- a/packages/container/libraries/strongly-typed/package.json
+++ b/packages/container/libraries/strongly-typed/package.json
@@ -1,0 +1,72 @@
+{
+  "author": "Alec Gibson",
+  "bugs": {
+    "url": "https://github.com/inversify/monorepo/issues"
+  },
+  "description": "InversifyJs strong type definitions",
+  "devDependencies": {
+    "@eslint/js": "9.13.0",
+    "@jest/globals": "29.7.0",
+    "@types/node": "20.17.4",
+    "@typescript-eslint/eslint-plugin": "8.12.2",
+    "@typescript-eslint/parser": "8.12.2",
+    "eslint": "9.13.0",
+    "jest": "29.7.0",
+    "prettier": "3.3.3",
+    "rimraf": "6.0.1",
+    "ts-jest": "29.2.5",
+    "ts-loader": "9.5.1",
+    "typescript": "5.6.3"
+  },
+  "devEngines": {
+    "node": "^20.18.0",
+    "pnpm": "^9.12.1"
+  },
+  "homepage": "https://inversify.io",
+  "keywords": [
+    "dependency injection",
+    "dependency inversion",
+    "di",
+    "inversion of control container",
+    "ioc",
+    "javascript",
+    "node",
+    "typescript"
+  ],
+  "license": "MIT",
+  "main": "lib/cjs/index.js",
+  "module": "lib/esm/index.js",
+  "exports": {
+    ".": {
+      "import": "./lib/esm/index.js",
+      "require": "./lib/cjs/index.js"
+    }
+  },
+  "name": "@inversifyjs/strongly-typed",
+  "peerDependencies": {
+    "inversify": ">=6.0.3"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/inversify/monorepo.git"
+  },
+  "scripts": {
+    "build": "pnpm run build:cjs && pnpm run build:esm",
+    "build:cjs": "tsc --build tsconfig.cjs.json && pnpm exec foundation-ts-package-cjs ./lib/cjs",
+    "build:esm": "webpack && pnpm exec foundation-ts-package-esm ./lib/esm",
+    "build:clean": "rimraf lib",
+    "format": "prettier --write ./src/**/*.ts",
+    "lint": "eslint ./src",
+    "prebuild": "pnpm run build:clean",
+    "test": "jest --config=jest.config.mjs --runInBand",
+    "test:js": "jest --config=jest.js.config.mjs --runInBand",
+    "test:js:coverage": "pnpm run test:unit:js --coverage",
+    "test:uncommitted": "pnpm run test --changedSince=HEAD",
+    "test:unit:js": "pnpm run test:js --selectProjects Unit"
+  },
+  "sideEffects": false,
+  "version": "1.0.0"
+}
--- a/packages/container/libraries/strongly-typed/src/container.spec.ts
+++ b/packages/container/libraries/strongly-typed/src/container.spec.ts
@@ -1,0 +1,206 @@
+/* eslint-disable @typescript-eslint/typedef */
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+/* eslint-disable jest/expect-expect */
+import { beforeEach, describe, expect, it } from '@jest/globals';
+
+import 'reflect-metadata';
+
+import { Container, injectable } from 'inversify';
+
+import { TypedContainer } from './index';
+
+describe('interfaces', () => {
+  @injectable()
+  class Foo {
+    public foo: string = '';
+  }
+
+  @injectable()
+  class Bar {
+    public bar: string = '';
+  }
+
+  describe('Container', () => {
+    let foo: Foo;
+    let foos: Foo[];
+
+    beforeEach(() => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      foo;
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      foos;
+    });
+
+    describe('no binding map', () => {
+      let container: TypedContainer;
+
+      beforeEach(() => {
+        container = new Container() as TypedContainer;
+      });
+
+      describe('bind()', () => {
+        it('binds without a type argument', () => {
+          container.bind('foo').to(Foo);
+          container.bind(Foo).to(Foo);
+        });
+
+        it('checks bindings with an explicit type argument', () => {
+          container.bind<Foo>('foo').to(Foo);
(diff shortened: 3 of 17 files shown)


packages/container/libraries/strongly-typed/src/container.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+      foo;
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      foos;
+    });
+
+    describe('no binding map', () => {
+      let container: TypedContainer;
+
+      beforeEach(() => {
+        container = new Container() as TypedContainer;
```
alecgibson, line 37: This type assertion is currently **required**. That is, you **can't** just write:

```ts
let container: TypedContainer;
container = new Container();
```

...because TypeScript considers the two types incompatible. This makes me feel a little uneasy, but I think it's unavoidable without touching upstream types.

packages/container/libraries/strongly-typed/src/container.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+        });
+
+        it('gets all', () => {
+          foos = container.getAll('foo');
+          // @ts-expect-error :: can't assign Bar to Foo
+          foos = container.getAll('bar');
+        });
+      });
+
+      describe('ancestry', () => {
```
alecgibson, line 131: NB: I've improved this since writing https://github.com/inversify/InversifyJS/pull/1575

packages/container/libraries/strongly-typed/src/container.ts
```diff
@@ -0,0 +1,197 @@
...
+type First<T extends any[]> = T extends [infer TFirst, ...any[]]
+  ? TFirst
+  : never;
+
+type AllButFirst<T extends any[]> = T extends [any, ...infer TRest]
+  ? TRest
+  : never;
+
+interface ContainerOverrides<
+  T extends BindingMap = any,
```
alecgibson, line 35: Now that these type definitions are opt-in, we *could* potentially make this a **required** type definition (otherwise why are you even using this interface), which might simplify types?

On the other hand, I quite like that this is vaguely backwards-compatible still, and possibly helps if you want to allow weird hierarchies involving children/parents:

```ts
const parent = new Container() as TypedContainer<ParentMap>;
const child = parent.createChild() // can create without binding map
// ...but this is still strongly typed despite `child` not having a mapping
child.parent.get('...') 
```

packages/container/libraries/strongly-typed/src/container.ts
```diff
@@ -0,0 +1,197 @@
...
+  TKey extends MappedServiceIdentifier<T>,
+>(
+  serviceIdentifier: TKey,
+) => boolean;
+
+export type TypedContainer<
+  T extends BindingMap = any,
+  TAncestors extends BindingMap[] = any[],
+> = ContainerOverrides<T, TAncestors> &
+  Omit<interfaces.Container, keyof ContainerOverrides>;
```
alecgibson, line 197: May want to discuss this. At the moment, I've deliberately omitted anything from the interface that doesn't have a type override.

This means we automatically get any new members that are added, which may be good for maintainability, although maybe it's a little "magical". There's an argument to be made for explicitly overriding things being clearer.

packages/container/libraries/strongly-typed/src/container.spec.ts
alecgibson: As discussed, we'll give consumers a couple of options to use this library:

```ts
container = new Container() as TypedContainer<Mapping>;
container = wrapContainer<Mapping>(new Container());
container = new TypedContainer<Mapping>();
```

packages/container/libraries/strongly-typed/src/container.ts
alecgibson: As discussed, we'll keep this as optional:

- makes it easier to move into `core` later if we want to
- probably works nicer with hierarchies
- more permissive, which is generally a bit nicer for consumers

packages/container/libraries/strongly-typed/src/container.ts
alecgibson: As discussed, will keep it like this to make sure we get new upstream members.

packages/container/libraries/strongly-typed/src/container.ts
```diff
@@ -0,0 +1,197 @@
...
+    value: unknown,
+  ) => TBound[];
+  getAllNamed: <
+    TBound extends ContainerBinding<T, TKey>,
+    TKey extends MappedServiceIdentifier<T> = any,
+  >(
+    serviceIdentifier: TKey,
+    named: PropertyKey,
+  ) => TBound[];
+  getAsync: <
```
alecgibson, line 82: Need to change the types for `get()` to **not** allow any mapped values that return a `Promise`, and **only** allow these bindings to be accessed by the async methods

packages/container/libraries/strongly-typed/src/container.spec.ts
alecgibson: Actually @notaphplover I've thought about it some more, and I *don't* think we should have the wrapper method (in the interests of keeping the API footprint small).

I think it offers no benefits over the other two methods, and instead pulls in the worst things about them. It's just performing a type assertion under the hood, but this is now less transparent. Since the function requires a non-`type` import, it also won't tree shake this dependency out of the final build.

notaphplover: Let's forget about the wraper then

packages/container/libraries/strongly-typed/package.json
```diff
@@ -0,0 +1,73 @@
...
+    ".": {
+      "import": "./lib/esm/index.js",
+      "require": "./lib/cjs/index.js"
+    }
+  },
+  "name": "@inversifyjs/strongly-typed",
+  "os": [
+    "darwin",
+    "linux"
+  ],
```
notaphplover, line 47:
```suggestion
```
Let's remove this. Windows users should use this library as well (I already change this in the other packages)

packages/container/libraries/strongly-typed/src/container.ts
```diff
@@ -0,0 +1,210 @@
...
+export type TypedContainer<
+  T extends BindingMap = any,
+  TAncestors extends BindingMap[] = any[],
+> = ContainerOverrides<T, TAncestors> &
+  Omit<interfaces.Container, keyof ContainerOverrides>;
+
+// eslint-disable-next-line @typescript-eslint/naming-convention, @typescript-eslint/no-unsafe-assignment
+export const TypedContainer: {
+  new <T extends BindingMap = any, TAncestors extends BindingMap[] = any[]>(
+    ...args: any[]
```
notaphplover, line 208:
```suggestion
    containerOptions?: interfaces.ContainerOptions,
```
Let's keep the original constructor params

packages/container/libraries/strongly-typed/src/container.ts
```diff
@@ -0,0 +1,210 @@
...
+
+type IfAny<T, TYes, TNo> = 0 extends 1 & T ? TYes : TNo;
+
+type BindingMapProperty = string | symbol;
+type BindingMap = Record<BindingMapProperty, any>;
+type BindingMapKey<T extends BindingMap> = keyof T & BindingMapProperty;
+type MappedServiceIdentifier<T extends BindingMap> = IfAny<
+  T,
+  interfaces.ServiceIdentifier,
+  BindingMapKey<T>
```
notaphplover, line 13:
```suggestion
  keyof T
```
I managed to compile the source code with this change :)

packages/container/libraries/strongly-typed/src/container.ts
```diff
@@ -0,0 +1,210 @@
...
+>;
+type ContainerBinding<
+  T extends BindingMap,
```
notaphplover, line 16:
```suggestion
  TBindingMap extends BindingMap,
```

packages/container/libraries/strongly-typed/README.md
```diff
@@ -0,0 +1,59 @@
+[![Test coverage](https://codecov.io/gh/inversify/monorepo/branch/main/graph/badge.svg?flag=%40inversifyjs%2Fstrongly-typed)](https://codecov.io/gh/inversify/monorepo/branch/main/graph/badge.svg?flag=%40inversifyjs%2Fstrongly-typed)
```
notaphplover, line 1: I'm missing a section talking about container hierarchies.

Since a child container overrides a parent container, so does their binding maps:

```ts
interface ParentBindingMap {
  foo: Foo;
  bar: Bar;
}

interface ChildBindingMap extends ParentBindingMap {
  bar: AlternativeBar;
  baz: Baz;
}
```

I think it would be great to include this detail in the docs.

packages/container/libraries/strongly-typed/src/container.ts
```diff
@@ -0,0 +1,210 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-unnecessary-type-parameters */
+import { Container, type interfaces } from 'inversify';
+
+type IfAny<T, TYes, TNo> = 0 extends 1 & T ? TYes : TNo;
+
+type BindingMapProperty = string | symbol;
+type BindingMap = Record<BindingMapProperty, any>;
```
notaphplover, line 8: Should we protect from async type mappings such as:

```ts
interface MyBindingMap {
  foo: Promise<Foo>;
}
```

I think we should add this to the docs or do some type mapping such as:

```ts
AwaitedBindingMap<TBindingMap extends BindingMap> = {
  [TKey in keyof TBindingMap]: Awaited<TBindingMap[TKey]>
}
```
This way we could receive a `BindingMap` and internally apply `AwaitedBindingMap<BindingMap>` to the `TypeContainer`.

notaphplover (changes requested):
It looks great.

I added some suggestions, what do you think about them?

packages/container/libraries/strongly-typed/README.md
alecgibson: Good point! I've now also strengthened the type signature of `createChild()` to enforce this, too. I've made it so that the child's binding map must extend the parent's binding map. Is this conceptually correct? ie in your example, is `AlternativeBar` meant to be compatible with `Bar`? (ie could they both be abstracted to implement some `CommonBar`)?

packages/container/libraries/strongly-typed/src/container.ts
alecgibson: I'm not sure I follow? I thought we allowed promises, and just enforced that they had to be accessed with `getAsync()`? Basically this test case: https://github.com/inversify/monorepo/pull/59/files#diff-9892fb8135f8ba0c3d8719dc0d16884330ab5a7651ac09aa65ffe97712f6784eR121-R128

I'm also updated the `README` to talk about `Promise`s.

packages/container/libraries/strongly-typed/README.md
notaphplover: `AlternativeBar` can be anything, related or unrelated to `Bar`. Maybe my example was sort of misleading :sweat_smile:

packages/container/libraries/strongly-typed/src/container.ts
notaphplover: You're right, we agreed on that. What should we do about the following update in your test:

```ts
it('gets a promise', async () => {
  // @ts-expect-error :: can't call get() to get Promise
  expect(() => container.get('asyncNumber')).toThrow(
    'it has asynchronous dependencies',
  );
  const n: Promise<number> = container.getAsync('asyncNumber');
  expect(await n).toBe(1);
});
```
It complains with the following error:

```
Type 'Promise<Promise<number>>' is not assignable to type 'Promise<number>'.
  Type 'Promise<number>' is not assignable to type 'number'.
```

But we do expect a `Promise<number>` instead of a `Promise<Promise<number>>`.

notaphplover (changes requested):
It looks great!

I think we're almost done, we just need to handle this scenario and we're ready to merge it:

```ts
it('gets a promise', async () => {
  // @ts-expect-error :: can't call get() to get Promise
  expect(() => container.get('asyncNumber')).toThrow(
    'it has asynchronous dependencies',
  );
  const n: Promise<number> = container.getAsync('asyncNumber'); // tsc complains
  expect(await n).toBe(1);
});

```

packages/container/libraries/strongly-typed/README.md
alecgibson: Oh okay in that case my typings are too strong. Will adjust.

alecgibson: Have actually readjusted this to automatically generate the correct binding map from a partial child map, because I think that's a friendlier API. Let me know if you disagree?

Have updated the readme, but copying here for discussion if needed:


> If the child is strongly-typed, its binding will automatically be adjusted to include the parent types:

> ```ts
> const parent = new TypedContainer<{foo: Foo}>();
> const child = parent.createChild<{bar: Bar}>();
> child.get('bar') // ok
> child.get('foo') // ok
> ```

> A child may also override a parent's bindings with a completely unrelated type:

> ```ts
> const parent = new TypedContainer<{foo: Foo}>();
> const child = parent.createChild<{foo: Bar}>();
> const resolved: Bar = child.get('foo');
> ```

ffMathy:
This is amazing.

Not sure why this is not offered by default. A lot of people are not going to find this when it's an extension, and TypeScript is almost used everywhere at this point.

Can you elaborate on the thoughts of not having this in the main project? The generic type could just default to "any" or something similar.

notaphplover:
Hey @ffMathy, it's a long story. I could resume it in these bullet points:
- I really want to avoid adding complexity in the main repository as much as possible until the monorepo migration is ready.
- This approach is amazing, but the logical enhancements around it suggests a config based bindings which are not exactly aligned with the code driven bindings original design. Keeping this out of the main repo allows us to carefully see how it evolves without introducing a feature in the main repo.
- The approach is practical for most cases, but it feels incomplete when some advanced tools like `middlewares` are involved. Keeping it out of the repo allows this type system to not to handle every impossible edge case you might face when using `inversify`. As a plugin, the plugin docs might perfectly have a `scope` section in which known limitations are ellaborated.

As said, this is just a summary.

Regarding visibility, I will update docs to give the visibility it deserves :smiley:

packages/container/libraries/strongly-typed/src/container.ts
```diff
@@ -0,0 +1,209 @@
...
+>;
+type ContainerBinding<
+  TBindingMap extends BindingMap,
+  TKey extends MappedServiceIdentifier<TBindingMap> = any,
+> = TKey extends keyof TBindingMap
+  ? TBindingMap[TKey]
+  : TKey extends interfaces.Newable<infer C>
+    ? C
+    : TKey extends interfaces.Abstract<infer D>
+      ? D
```
notaphplover, line 22: I didn't see this in preview reviews..., sry. Ok, `interfaces.ServiceIdentifier<T>` no longer includes `interfaces.Abstract<T>`. Instead, it includes the expected type `Function`. Since `Abstract<T>` was just a trick to exploit the `any` type of a `Function`, what about this:
```suggestion
    : TKey extends Function
      ? any
```

notaphplover (commented):
Ok, I saw a last minute issue, adding the comment.

I was about to do this myself but, would you mind adding an ESM build the same way we're actually doing with other modules? Otherwise I'll do it after we merge this.

Thank you so much for your great job!

packages/container/libraries/strongly-typed/README.md
notaphplover: Love it

alecgibson:
@notaphplover without changing anything my build now fails locally:

```
❯ pnpm test

> @inversifyjs/strongly-typed@1.0.0 test /Users/alec/git/work/reedsy/inversify/monorepo/packages/container/libraries/strongly-typed
> jest --config=jest.config.mjs --runInBand

● Validation Error:

  Module ts-jest in the transform option was not found.
         <rootDir> is: /Users/alec/git/work/reedsy/inversify/monorepo/packages/container/libraries/strongly-typed

  Configuration Documentation:
  https://jestjs.io/docs/configuration

 ELIFECYCLE  Test failed. See above for more details.
```

Did we change something upstream that I need to adapt to?

The overly-aggressive commit hooks won't let me commit without fixing this.

notaphplover:
> Did we change something upstream that I need to adapt to?

Hey @alecgibson , everything seems to be fine in th CI environment and mine.
 
> The overly-aggressive commit hooks won't let me commit without fixing this.

git hooks are client side so, at the end, are optional. You can bypass then using the `--no-verify` option:

```bash
git commit --no-verify
```

packages/container/libraries/strongly-typed/src/container.ts
alecgibson: Okay not entirely sure I follow, but have done this anyway since it matches upstream.

alecgibson:
> You can bypass then using the --no-verify option

Thanks! Didn't know about this. (I've always found commit hooks a bit heavy-handed for my personal taste 🤷🏼 ).

I've made the requested changes (I think? I wasn't sure what you meant about the ESM build; I just copied the `webpack` config and `script` from `core`).

Want to also make sure you're happy with the location of this library in the folder structure?

ffMathy:
So exciting! Great work everyone! 😍

notaphplover (approved):

notaphplover (approved):

ffMathy:
The package seems to not be public somehow?

If I go to https://www.npmjs.com/package/@inversifyjs/strongly-typed, I am taken to a login page.

notaphplover:
Hey @ffMathy, the package is not ready to be published. Additional PR are required to provide with `inject` decorators relying on a `BindingMap`

ffMathy:
Okay. Is this being worked on and tracked anywhere? And if not, is that something I can work on?

alecgibson:
For what it's worth I personally consider this `TypedContainer` feature "useful enough" as its own thing to release it without the strongly-typed `@inject()` decorators, which would be a minor-versioned, non-breaking update anyway. I'll work on that hopefully this week when I get some time.

alecgibson:
`inject` typings now raised here: https://github.com/inversify/monorepo/pull/114

PR #114, opened 2024-11-19
feat(strongly-typed): inject decorator type

alecgibson:
This change builds on the work of the [`TypedContainer`][1] by adding type definitions for a strongly-typed `inject` decorator.

[1]: https://github.com/inversify/monorepo/pull/59


packages/container/libraries/strongly-typed/README.md
```diff
@@ -108,3 +146,39 @@ const parent = new TypedContainer<{foo: Foo}>();
...
+
+### Confusing compiler errors for constructor injection
+
+There's no custom compiler errors for TypeScript, so we can't signal particularly usefully that a constructor parameter is wrong, apart from the fact that the compilation fails.
+
+The error will look something like:
+
+```
+Unable to resolve signature of parameter decorator when called as an expression.
+  Argument of type '2' is not assignable to parameter of type 'undefined'. (ts1239)
```
alecgibson, line 179: @notaphplover I couldn't really get a very useful compiler message either. I went with this, which is a bit different to your proposal, because:

- the type definition was simpler
- the error message at least flags that there's something wrong with argument `2`, which is sort of meaningful

I don't think there's much we can do without this: https://github.com/microsoft/TypeScript/issues/23689

packages/container/libraries/strongly-typed/src/inject.ts
```diff
@@ -0,0 +1,24 @@
...
+        : never
+      : never
+    : unknown,
+) => void;
+
+export type TypedInject<TBindingMap extends BindingMap> = <
+  TKey extends keyof TBindingMap,
+>(
+  identifier: TKey,
+) => TypedDecorator<TBindingMap[TKey]>;
```
notaphplover, line 24:
```suggestion
) => TypedDecorator<Awaited<TBindingMap[TKey]>>;
```

Container injects the awaited value of a promise.

packages/container/libraries/strongly-typed/src/inject.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  @injectable()
+  class Bar {
+    public bar: string = '';
+  }
+
+  interface BindingMap {
+    foo: Foo;
+    bar: Bar;
+    asyncNumber: Promise<number>;
```
notaphplover, line 25: We should add a test involving this field

packages/container/libraries/strongly-typed/README.md
notaphplover: Agree, the error is way better than the one experienced with the approach I suggested.

Regarding the ts limitation, I'm afraid so.

notaphplover (changes requested):
The PR looks great! We just need to handle async mapped types properly and add a test case for it.

packages/container/libraries/strongly-typed/src/inject.spec.ts
alecgibson: Ha added this and completely forgot to test it 🤦🏼

packages/container/libraries/strongly-typed/src/inject.ts
alecgibson: (I don't think I knew that was how this worked 😅 I think I always assumed `@inject()` needed synchronous dependencies)

notaphplover (approved):
Lgtm!

PR #159, opened 2024-12-04
feat(strongly-typed): add `@multiInject()` types

alecgibson:


notaphplover (commented):
Great job! Can you add a changeset with your changes?

You should be able to do it via `pnpm changeset`

alecgibson:
@notaphplover I ran `pnpm changeset` and it added a markdown file in the root `.changeset/` directory. Is that what you wanted? (I couldn't see any evidence of this being done for any other packages, so was unsure if I was doing it wrong...?)

notaphplover:
> @notaphplover I ran `pnpm changeset` and it added a markdown file in the root `.changeset/` directory. Is that what you wanted? (I couldn't see any evidence of this being done for any other packages, so was unsure if I was doing it wrong...?)

Hey @alecgibson , yeah, we use changesets to register our changes. When a new release is cut, changelogs are automatically generated from those changesets and are removed

notaphplover (approved):

PR #416, opened 2025-02-13
Add speedups to benchmark results

Adrianmjim:
### Added
- Add speedups to benchmark results

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,6 +3,12 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 on:
+  pull_request_target:
+    types:
+      - closed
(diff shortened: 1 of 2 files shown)


notaphplover:
Hey @Adrianmjim, I see there's an unrelated issue with permissions. This issue never happened before, probably due to the fact GH tokens issued on PR authored by me had permissions.

Can you update the `run-container-benchmarks` at `build.yaml` with `issue` and `pull-requests` write [premissions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token)? I'm afraid if I do the change I might accidentally solve it thanks to my tokens instead of solving it with a code change.

adrianmjim:
Hi @notaphplover,

Understood, I updated the branch with requested permissions

PR #430, opened 2025-02-21
Add http-core package

Adrianmjim:
### Added
- Create http-core package
- Add http-core basic models
- Add http-core request method decoratos
- Add http-core controller decorator
- Add http-core request param decorators
- Add InversifyHttpAdapter


packages/http/libraries/core/README.md
```diff
@@ -0,0 +1,6 @@
+[![Test coverage](https://codecov.io/gh/inversify/monorepo/branch/main/graph/badge.svg?flag=%40inversifyjs%2Fcommon)](https://codecov.io/gh/inversify/monorepo/branch/main/graph/badge.svg?flag=%40inversifyjs%2Fcommon)
+[![npm version](https://img.shields.io/github/package-json/v/inversify/monorepo?filename=packages%2Fcontainer%2Flibraries%2Fcommon%2Fpackage.json&style=plastic)](https://www.npmjs.com/package/@inversifyjs/common)
```
notaphplover, line 2: Let's update these links to point to the right libraries

packages/http/libraries/core/src/adapter/InversifyHttpAdapter.ts
```diff
@@ -0,0 +1,209 @@
...
+    this.#registerControllers();
+  }
+
+  #registerControllers(): void {
+    const controllerMetadataList: ControllerMetadata[] | undefined =
+      getReflectMetadata(Reflect, METADATA_KEY.controller);
+
+    if (controllerMetadataList === undefined) {
+      throw new Error('No controllers found');
+    }
```
notaphplover, line 32: I am not sure we should throw here. if we do, let's have custom errors the same way the container core library has :)

packages/http/libraries/core/src/adapter/InversifyHttpAdapter.ts
```diff
@@ -0,0 +1,209 @@
...
+        );
+
+        const controller: { [key: string | symbol]: () => unknown } =
+          this.#container.get(controllerMetadata.target);
+
+        return {
+          handler: async (
+            req: TRequest,
+            res: TResponse,
+            next: TNextFunction,
```
notaphplover, line 79: I wonder if this approach is generic enough, I believe so but maybe we should double check

packages/http/libraries/core/src/models/RouterParams.ts
```diff
@@ -0,0 +1,9 @@
+import { ControllerMethodMetadata } from './ControllerMethodMetadata';
+import { RequestHandler } from './RequestHandler';
+
+export type RouterParams<TRequest, TResponse, TNextFunction> = Omit<
+  ControllerMethodMetadata,
+  'methodKey'
+> & {
+  handler: RequestHandler<TRequest, TResponse, TNextFunction>;
+};
```
notaphplover, line 9: I would prefer to avoid `Omit` as much as possible.

packages/http/libraries/core/src/models/RequestMethodType.ts
```diff
@@ -0,0 +1,10 @@
+export enum RequestMethodType {
+  GET = 'get',
+  POST = 'post',
+  PUT = 'put',
+  DELETE = 'delete',
+  PATCH = 'patch',
+  OPTIONS = 'options',
+  //HEAD = 'head',
```
notaphplover, line 8: Let's include head requests as well

packages/http/libraries/core/src/adapter/InversifyHttpAdapter.ts
```diff
@@ -0,0 +1,209 @@
...
+          case RequestMethodParameterType.COOKIES: {
+            return this._getCookies(
+              request,
+              controllerMethodParameterMetadata.parameterName,
+            );
+          }
+          case RequestMethodParameterType.NEXT: {
+            return next;
+          }
+          case RequestMethodParameterType.PRINCIPAL: {
```
notaphplover, line 162: I'm curious about this :)

packages/http/libraries/core/src/decorators/Delete.ts
```diff
@@ -0,0 +1,6 @@
+import { RequestMethodType } from '../models/RequestMethodType';
+import { requestMethod } from './RequestMethod';
+
+export function delet(path?: string): MethodDecorator {
```
notaphplover, line 4:
```suggestion
export function delete(path?: string): MethodDecorator {
```

notaphplover (changes requested):
It looks nice! I really like the abstraction. I added some minor details to polish.

packages/http/libraries/core/src/adapter/InversifyHttpAdapter.ts
Adrianmjim: This functionality returns the user from the request in the current library. We currently do not have an authentication service, which is why it is disabled.

Anyways, if we enable authentication services, do we want to keep this functionality with the same name and usage? I'm not sure.

packages/http/libraries/core/src/decorators/Delete.ts
Adrianmjim: We have a problem here: `delete` is a reserved word in ts

packages/http/libraries/core/src/models/RequestMethodType.ts
Adrianmjim: I have to find a way to include head function to hono driver. Anyways, I include it.

packages/http/libraries/core/src/adapter/InversifyHttpAdapter.ts
Adrianmjim: It's not, but imho this is a first iteration

packages/http/libraries/core/README.md
Adrianmjim: Fixed

packages/http/libraries/core/src/adapter/InversifyHttpAdapter.ts
Adrianmjim: I refactored this error in that way

packages/http/libraries/core/src/models/RouterParams.ts
Adrianmjim: Fixed

packages/http/libraries/core/src/models/RequestMethodType.ts
notaphplover: Yeah, If the adapter cannot implement a feature we can trow a custom error

notaphplover (commented):
It looks great, let's fix CI issues and I think we can merge

packages/http/libraries/core/src/models/RequestMethodType.ts
Adrianmjim: HEAD value added

packages/http/libraries/core/src/decorators/Delete.ts
Adrianmjim: Fixed

notaphplover (approved):

PR #437, opened 2025-02-24
Refactor controller method parameter metadata

Adrianmjim:
## Refactor
- refactor controller method parameter metadata to get assigned at controller function metadata


notaphplover (changes requested):
It looks good, let's add some additional checks and we can merge it

packages/http/libraries/core/src/http/decorators/RequestParam.ts
```diff
@@ -16,14 +20,21 @@ export function requestParam(
...
-      | {
-          [key: string]: ControllerMethodParameterMetadata[];
-        }
-      | undefined = getReflectMetadata(
-      target.constructor,
-      controllerMethodParameterMetadataReflectKey,
-    );
+    const controllerFunction: ControllerFunction | undefined = (
+      target as Controller
+    )[key as string];
```
notaphplover, line 25: Let's avoid accesing if key is undefined:

```suggestion
    const controllerFunction: ControllerFunction | undefined = key === undefined ? undefined : (
      target as Controller
    )[key];
```

notaphplover (changes requested):
Almost there! I saw a casting I believe we should avoid

notaphplover (approved):
lgtm!

PR #438, opened 2025-02-24
Add status code decorator

Adrianmjim:
### Added
- Add statusCode decorator

### Motivation
Continuing with industry standards, it’s interesting to have a decorator that allows changing the status code of the request without it being handled in the body of the function.


packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -188,19 +195,20 @@ export abstract class InversifyHttpAdapter<
     request: TRequest,
     response: TResponse,
     value: ControllerResponse,
+    statusCode?: HttpStatusCode,
   ): unknown {
     let body: object | string | number | boolean | undefined = undefined;
-    let statusCode: HttpStatusCode | undefined = undefined;
+    let httpStatusCode: HttpStatusCode | undefined = statusCode;
 
     if (value instanceof HttpResponse) {
```
notaphplover, line 203: My bad for not seeing this before, never use `instanceof` on non native types, a user might end up with multiple versions of the core. Instead, consider an `is` static method the same way `AppError` does

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -188,19 +195,20 @@ export abstract class InversifyHttpAdapter<
...
       body = value;
     }
 
-    if (statusCode !== undefined) {
-      this._setStatus(request, response, statusCode);
+    if (httpStatusCode !== undefined) {
+      this._setStatus(request, response, httpStatusCode);
     }
 
     if (typeof body === 'string') {
```
notaphplover, line 214: Maybe we should rely on the [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) header or delegate in the integration to decide whether or not string (text/plain mime type) or json (application/json mime) are used

packages/http/libraries/core/src/http/models/MetadataKey.ts
```diff
@@ -2,8 +2,11 @@ export const METADATA_KEY: {
   controller: string;
```
notaphplover, line 2: Let's rely on multiple constants instead of in a single object. Consider `@inversifyjs/core` reflect keys as reference

notaphplover (approved):
Looks good. I added some unrelated suggestions you can apply in other PR :)

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
Adrianmjim: Fixed

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
Adrianmjim: That a good design question. We have 2 ways:
- Delegate to http drive
- Get agnostic response

Second option is the current implementation. idk what option is better, for me it's just a approach

packages/http/libraries/core/src/http/models/MetadataKey.ts
Adrianmjim: Fixed

notaphplover (approved):

PR #443, opened 2025-02-25
Add get controller by name feature

Adrianmjim:
### Added
- Add get controller by name feature to InversifyHttpAdapter

### Motivation
Currently, inversify container is able to inject multiple objects by name. With this feature, It's posible to match controllers injected by name with his class representation.


notaphplover (approved):
lgtm

PR #444, opened 2025-02-25
Add middleware support

Adrianmjim:
### Added
- Add middleware support to RouteParams
- Add middleware prototype
- Add applyMiddleware decorator
- Add middleware build logic to InversifyHttpAdapter


packages/http/libraries/core/src/http/decorators/ApplyMiddleware.ts
```diff
@@ -0,0 +1,52 @@
...
+      );
+    } else {
+      middlewareMetadataList = getReflectMetadata(
+        descriptor.value as ControllerFunction,
+        controllerMethodMiddlewareMetadataReflectKey,
+      );
+    }
+
+    if (middlewareMetadataList !== undefined) {
+      middlewareList.push(...middlewareMetadataList);
```
notaphplover, line 33: I'm a little bit concerned about middleware order. Given two apply middleware calls, one at as class decorator and another one as method decorator, can we guarantee a deterministic order? The implementation suggests class level middlewares should be pushed first to `middlewareList`, but maybe we should double check it.

notaphplover (commented):
The code looks good. I have some doubts regarding middleware preference. We should probably add unit tests as well

packages/http/libraries/core/src/http/decorators/ApplyMiddleware.ts
Adrianmjim: The goal is to ensure that middleware for the endpoint and the controller are applied separately. In such a way that the class middlewares can be applied to the router and the endpoint middlewares exclusively to the endpoint.

Adrianmjim: Oh, now I know what You mean. It was a mistake. I fixed applyMiddleware decorator. Reanalyze pls.

notaphplover: I had it a look. Nice catch, but my point is a little bit different. Consider the following example:

```ts
function decorator(...params: any[]) {
    return (...params: any[]) => undefined
}

@decorator('call1')
class A {
    @decorator('call2')
    public b(): void {
        return undefined
    }
}
```

This code is transpiled as follows:

```js
"use strict";
var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
    var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
    if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);
    else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;
    return c > 3 && r && Object.defineProperty(target, key, r), r;
};
var __metadata = (this && this.__metadata) || function (k, v) {
    if (typeof Reflect === "object" && typeof Reflect.metadata === "function") return Reflect.metadata(k, v);
};
function decorator(...params) {
    return (...params) => undefined;
}
let A = class A {
    b() {
        return undefined;
    }
};
__decorate([
    decorator('call2'),
    __metadata("design:type", Function),
    __metadata("design:paramtypes", []),
    __metadata("design:returntype", void 0)
], A.prototype, "b", null);
A = __decorate([
    decorator('call1')
], A);
```

Keeping this in mind, it seems `applyMiddleware` calls as `MethodDecorator` take precedence to calls as `ClassDecorator`. Are we ok with this? I think it makes sense. Let's remember to specify this in the docs in the future.

notaphplover (approved):
lgtm!

PR #461, opened 2025-03-01
Add injectable call to controller decorator

Adrianmjim:
### Added
- Add injectable call to controller decorator logic

### Motivation
With new version of inversify http integration, controllers are considered object of injection system. We need to add injectable call to controller logic.

### DIFF
--- a/packages/http/libraries/core/src/http/decorators/Controller.ts
+++ b/packages/http/libraries/core/src/http/decorators/Controller.ts
@@ -10,14 +10,14 @@ import { ControllerOptions } from '../models/ControllerOptions';
 
 export function controller(
   pathOrOptions?: string | ControllerOptions,
-  scope?: BindingScope,
 ): ClassDecorator {
   return (target: NewableFunction): void => {
-    injectable(scope)(target);
     const controllerMetadata: ControllerMetadata = {
       path: '/',
       target,
     };
+
+    let scope: BindingScope | undefined = undefined;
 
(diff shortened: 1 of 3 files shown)


packages/http/libraries/core/src/http/decorators/Controller.ts
```diff
@@ -2,15 +2,18 @@ import {
...
 } from '@inversifyjs/reflect-metadata-utils';
+import { BindingScope, injectable } from 'inversify';
 
 import { controllerMetadataReflectKey } from '../../reflectMetadata/data/controllerMetadataReflectKey';
 import { ControllerMetadata } from '../models/ControllerMetadata';
 import { ControllerOptions } from '../models/ControllerOptions';
 
 export function controller(
   pathOrOptions?: string | ControllerOptions,
+  scope?: BindingScope,
```
notaphplover, line 13: Does it make sense to be part of  `ControllerOptions`?. I know the implementation won't be as clean as the current one but I think the interface makes more sense that way.

notaphplover (approved):
The code looks good, I have a suggestion regarding the `scope` argument.

packages/http/libraries/core/src/http/decorators/Controller.ts
Adrianmjim: Yeah, it makes sense. i'm going to refactor it. No problem with new implementation

notaphplover (approved):

PR #462, opened 2025-03-01
Vitest Migration

Adrianmjim:
## Added
- Add vitest-config package
- Add eslint-vitest-plugin config

## Refactor
- Migrate binding-decorators package to vitest
- Migrate common package to vitest
- Migrate container package to vitest
- Migrate core package to vitest
- Migrate strongly-typed package to vitest
- Migrate binding-decorators-code-examples package to vitest
- Migrate code-examples package to vitest
- Migrate prototype-utils package to vitest
- Migrate reflect-metadata-utils package to vitest
- Migrate http-core package to vitest


packages/foundation/tools/eslint-config/index.js
```diff
@@ -186,6 +187,27 @@ export default tseslint.config(
...
+    files: ['**/*.spec.ts'],
+    plugins: {
+      vitest,
+    },
+    rules: {
+      ...vitest.configs.recommended.rules,
+      ...vitest.configs.all.rules,
+      '@typescript-eslint/unbound-method': 'off',
+      '@typescript-eslint/no-confusing-void-expression': 'off',
+      '@typescript-eslint/no-empty-function': 'off',
```
notaphplover, line 200: Why is this rule disabled?

packages/foundation/tools/eslint-config/index.js
```diff
@@ -186,6 +187,27 @@ export default tseslint.config(
...
+      '@typescript-eslint/no-magic-numbers': 'off',
+      '@typescript-eslint/no-non-null-asserted-optional-chain': 'off',
```
notaphplover, line 202: Let's avoid disabling this rule in favor of castings if required

packages/container/libraries/common/vitest.js.workspace.mjs
```diff
@@ -0,0 +1,3 @@
+import { jsWorkspaceConfig } from '@inversifyjs/foundation-vitest-config';
+
+export default jsWorkspaceConfig;
```
notaphplover, line 3: Let's get rid of this, I honestly believe it's not worth to target compiled tests.

notaphplover (changes requested):
Nice, I added some comments but overall it looks good :)

packages/foundation/tools/eslint-config/index.js
Adrianmjim: Sorry, I just copied from another project, I can't remember it xD. I will remove it

packages/foundation/tools/eslint-config/index.js
```diff
@@ -186,6 +187,26 @@ export default tseslint.config(
...
+      '@typescript-eslint/no-magic-numbers': 'off',
+      '@typescript-eslint/no-unsafe-assignment': 'off',
+      'vitest/consistent-test-filename': 'off',
+      'vitest/max-expects': 'off',
+      'vitest/no-hooks': 'off',
+      'vitest/prefer-expect-assertions': 'off',
+      'vitest/prefer-strict-equal': 'error',
+      'vitest/valid-title': 'off',
+    },
+  },
```
notaphplover, line 209: I think these rules are causing trouble to non migrated test files. Should we migrate everything in this PR?

notaphplover (commented):
Ok, I see vitest linter rules are causing trouble to jest tests. Does it make sense to migrate everything in this PR? We can do some pairing if you want

PR #463, opened 2025-03-01
Avoid controller handler build when controller is not injected

Adrianmjim:
## Added
- Avoid controller handler build when controller is not injected


notaphplover (approved):
lgtm!

PR #466, opened 2025-03-03
🏷️ BREAKING CHANGE(strongly-typed): Update to `inversify@7`

alecgibson:


packages/container/libraries/strongly-typed/src/container.spec.ts
```diff
@@ -155,45 +154,21 @@ describe('interfaces', () => {
...
-          const child = container.createChild();
+          const child = new TypedContainer({ parent: container });
           child.bind('unknown').toConstantValue('unknown');
           expect(child.get('unknown')).toBe('unknown');
         });
 
-        it('automatically extends parent types', () => {
-          const child = container.createChild<{ childProp: string }>();
+        it('allows a TypedContainer parent', () => {
+          const child = new TypedContainer<{ childProp: string } & BindingMap>({
```
alecgibson, line 163: We can possibly make these types stronger/smarter again in future, but I'm not sure I have the time to work it out right now. For now, it's left to consumers to correctly combine the child and parent binding maps.

packages/container/libraries/strongly-typed/src/container.ts
```diff
@@ -1,22 +1,33 @@
...
+  BindToFluentSyntax,
+  ContainerOptions,
+  IsBoundOptions,
+} from '@inversifyjs/container';
+import { Container } from '@inversifyjs/container';
+import type {
+  BindingActivation,
+  BindingDeactivation,
+  GetOptions,
+} from '@inversifyjs/core';
```
notaphplover, line 14: I would hazard to say these types are exposed in `inversify@7`. I prefer to rely on `inversify@7` since `@inversify/*` dependencies are internal ones and major releases are released way more often.

notaphplover (changes requested):
Awesome job! Thank you so much. I just requested to rely on `inversify` types.

We've migrated to vitest so you are likelly to have conflicts when updating with the main branch. If you see it's too much work just ping me and I'll do it for you :)

packages/container/libraries/strongly-typed/src/container.ts
alecgibson: I originally tried to do this but I think I'm missing the `BindingActivation` and `BindingDeactivation` types.

alecgibson: I've rebased onto `main`. Let me know what you want me to do about these types. Options:

- expose them in `inversify` (probably best option? For this library at least)
- hand-roll the type to be compatible (fragile, but doesn't require a change to `inversify`)
- add a dependence on `@inversifyjs/core` (but you said you don't want to do that)
- other...?

notaphplover: I see, yeah, `BindingActivation` and `BindingDeactivation` must be exposed in `inversify`. I'll do it during the weekend.

notaphplover (approved):

PR #468, opened 2025-03-04
Add minimal logger to HttpAdapter

Adrianmjim:
### Added
- Add HttpAdapterOptions
- Add minimal logger for injected controllers


packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -29,9 +30,14 @@ export abstract class InversifyHttpAdapter<
   TNextFunction extends (err?: unknown) => void,
 > {
   readonly #container: Container;
+  readonly #httpAdapterOptions: HttpAdapterOptions;
```
notaphplover, line 33:
```suggestion
  readonly #httpAdapterOptions: Required<HttpAdapterOptions>;
```

I would love to avoid dealing with nullish internal properties, does it make sense? For now `Required` is good enough, but we can have a better type once we need it.

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -279,6 +296,20 @@ export abstract class InversifyHttpAdapter<
     return requestHandlerList;
   }
 
+  #printController(
+    controllerName: string,
+    path: string,
+    controllerMethodMetadataList: ControllerMethodMetadata[],
+  ): void {
+    console.log(`${controllerName} {${path}}:`);
```
notaphplover, line 304: I know I said yes to this approach, but let's make an abstraction with some logging levels. I think [winston](https://github.com/winstonjs/winston?tab=readme-ov-file#logging-levels) does a pretty good job in this regard.

What do you think about it?

notaphplover (changes requested):
Let's add some basic abstractions to express logging levels

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
Adrianmjim: I understand, but there will be situations where we want optional properties to have default behaviors, as is the case here. It seems to me a more user-friendly approach.

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
Adrianmjim: I agree, but I consider that a logging system can be large and critical enough to be out of scope in this PR.

notaphplover (approved):

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
notaphplover: Indeed, let's keep the interface for the user, but let's internally use another with non nullish values :)

Issue #475, opened 2025-03-09
[Feature, http]: Implement guards


notaphplover:
### 🚀 Feature Proposal

Guards are responsible to determine whether or not a request will be handled by a route handler.

```ts
interface Guard<TRequest> {
  handle(request: TRequest): boolean | Promise<boolean>;
}
```

Given a controller with route handlers, a guard can be applied to a single route or to every route of the controller. An `applyGuards` decorator is proposed to anotate the relevant metadata.

If `guard.handle` returns true, the request won't be blocked by this guard. Otherwise, a `403` (Forbidden) error is thrown. Users might decide to throw custom forbidden errors, which is fine.

### Lifecycle
Guards should be executed before of after middlewares?

### Motivation

Guards are a popular mechanism to block request and implement auth mechanisms.

### Example

```ts

class MyAwesomeGuard implements Guard<Request> {
  public handle(request: TRequest): boolean {
    return true;
  }
}

@controller('/foo')
class MyController {
  @applyGuards(MyAwesomeGuard)
  @post('/bar')
  createBar(@request() request: Request, @response() response: Response): Promise<Bar> {
    return new Bar();
  }
}
```

### Pitch

-

PR #476, opened 2025-03-09
Add guards

Adrianmjim:
### Added 
- Add basic Guard
- Add auth Guard
- Add Generic UserRequest
- Add useGuard decorator
- Add guard logic to InversifyHttpAdapter

### Motivation
See #475


packages/http/libraries/core/src/http/guard/Guard.ts
```diff
@@ -0,0 +1,36 @@
...
+import { HttpAdapter } from '../adapter/HttpAdapter';
+import { InversifyHttpAdapter } from '../adapter/InversifyHttpAdapter';
+import { Middleware } from '../models/Middleware';
+
+@injectable()
+export abstract class Guard<
+  TRequest,
+  TResponse,
+  TNextFunction extends () => void,
+> implements Middleware<TRequest, TResponse, TNextFunction>
```
notaphplover, line 12: Why does `Guard` extend `Middleware`? I think `Guard should not receive responses nor next functions. Consider  #475 as reference

packages/http/libraries/core/src/http/guard/AuthGuard.ts
```diff
@@ -0,0 +1,41 @@
+import { injectable } from 'inversify';
+
+import { UserRequest } from '../models/UserRequest';
+import { Guard } from './Guard';
+
+@injectable()
+export abstract class AuthGuard<
```
notaphplover, line 7: Let's avoid implementing authentication in this PR. Auth is a complex topic to tackle. I would love to have an abstraction that covers user and non user flows with different flows such as oauth and oidc.

notaphplover (changes requested):
Nice, I think auth is out of the scope of this PR though

packages/http/libraries/core/src/http/guard/Guard.ts
Adrianmjim: Here I am considering that a guard is a middleware with more logic. This way, we can treat guards as if they were middlewares with a specific order in the adapter. Anyway, I will try to simplify the guard's contract.

packages/http/libraries/core/src/http/guard/AuthGuard.ts
Adrianmjim: Okay, imho we should discuss this more thoroughly. By removing the auth, we are not complying with the current API.

notaphplover: It think its ok for now. I prefer to keep the PR lighter this way and implement the guard in further PR

notaphplover (approved):

notaphplover (approved):

PR #483, opened 2025-03-12
Fix controller handler build

Adrianmjim:
### Fixed
- Fix controller handler build to use injected controllers


notaphplover (approved):
lgtm, let's sync with main and merge :smiley:

notaphplover (approved):

PR #487, opened 2025-03-13
fix!: make `toResolvedValue` mostly typesafe

BurningEnlightenment:
`toResolvedValue` receives an array of typed ServiceIdentifiers, therefore we should ensure that the types of the requested services matches the type signature of the factory function.


packages/container/libraries/container/src/binding/models/BindingFluentSyntaxImplementation.spec.ts
```diff
@@ -518,6 +518,88 @@ describe(BindToFluentSyntaxImplementation.name, () => {
...
+        it('should call callback', () => {
+          expect(callbackMock).toHaveBeenCalledTimes(1);
+          expect(callbackMock).toHaveBeenCalledWith(expectedBinding);
+        });
+
+        it('should return expected result', () => {
+          expect(result).toBeInstanceOf(BindInWhenOnFluentSyntaxImplementation);
+        });
+      });
+    });
```
notaphplover, line 601: I do appreciate the test, but this file is for unit testing purposes :smile:. Mhmm I honestly don't know what to do here, testing a type's behavior is out of the scope of a unit test imo but I'm not sure what kind of test should cover that. Since we are checking `BindingFluentSyntax` type behavior, the test should probably be decoupled from the implementation.

I have something in mind: we can add a `BindingFluentSyntax.types.spec.ts` with test to check just the types. `// @ts-expect-error` and some [vite cool tools](https://vitest.dev/guide/testing-types.html) are probably the way to go. Wanna give it a try? Otherwise I'll write the test :smiley:

notaphplover (changes requested):
Love the idea :smiley: , let's do some minor changes and merge it.

Just one last minor detail: we are using [changesets](https://github.com/changesets/changesets) so please add the changeset. I would hazard to say extra type safety can be considered a fix in the types, so I would consider adding a patch changeset.

notaphplover:
I fixed some comments and added some tests

packages/container/libraries/container/src/binding/models/BindingFluentSyntax.spec-d.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  beforeAll(() => {
+    bindToFluentSyntaxMock = {
+      toResolvedValue: vitest.fn() as unknown,
+    } as Partial<Mocked<BindToFluentSyntax<unknown>>> as Mocked<
+      BindToFluentSyntax<unknown>
+    >;
+  });
+
+  describe('.toResolvedValue', () => {
```
notaphplover, line 29: Some tests are missing, we need to test `ResolvedValueMetadataInjectOptions` with `isOptional` and `isMultiple`. Spoiler alert: we need to update typings to fix that use case

notaphplover (changes requested):
Let's add some tests, polish `ResolvedValueMetadataInjectOptions` and we should be ready to merge

notaphplover (approved):

BurningEnlightenment:
@notaphplover sorry for ghosting, I'm currently short on free time.

notaphplover:
No worries @BurningEnlightenment, your contribution was, indeed, helpful. I'm glad we could add some type safety there. I hope you can have more free time soon :slightly_smiling_face:

PR #490, opened 2025-03-14
Update InversifyHttpAdapter to rely on container async calls

notaphplover:
### Changed
- Updated `InversifyHttpAdapter` to rely on `container.getAsync` calls.

### Motivation
- Async http services such as guards or middlewares with async dependencies crash unless `container.getAsync` is used.


packages/http/libraries/core/src/routerExplorer/RouterExplorer.ts
```diff
@@ -21,21 +21,28 @@ import { RouterExplorerControllerMetadata } from './model/RouterExplorerControll
...
+  #routerExplorerControllerMetadataList:
+    | RouterExplorerControllerMetadata[]
+    | undefined;
   readonly #container: Container;
 
   constructor(container: Container) {
     this.#container = container;
-
-    this.#routerExplorerControllerMetadataList =
-      this.#buildControllerMetadataList();
```
notaphplover, line 31: Trying to avoid logic in constructors for async flows cannot be included

Adrianmjim (approved):

PR #494, opened 2025-03-15
Add custom logger to HttpAdapterOptions

Adrianmjim:
### Added
- Add custom logger to HttpAdapterOptions

### Motivation
In some scenarios, users will want to pass a custom logger to the adapter to handle the adapter's logs in a custom way


notaphplover (approved):
:heart:

notaphplover (approved):

PR #495, opened 2025-03-15
Log errors when request handler fails

Adrianmjim:
### Added
- Log errors when request handler fails at InversifyHttpAdapter

### DIFF
--- a/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
+++ b/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
@@ -19,6 +19,8 @@ import { InternalServerErrorHttpResponse } from '../responses/error/InternalServ
 import { HttpResponse } from '../responses/HttpResponse';
 import { HttpStatusCode } from '../responses/HttpStatusCode';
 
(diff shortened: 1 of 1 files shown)


packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -133,7 +133,8 @@ export abstract class InversifyHttpAdapter<
         ]?.(...handlerParams);
 
         return this.#reply(req, res, value, statusCode);
-      } catch (_error: unknown) {
+      } catch (error: unknown) {
+        this.#logger.error((error as Error).stack ?? (error as Error).message);
```
notaphplover, line 137: `error` might not be an `Error`. JS allow throwing any value like `throw 3`. A typeguard should be added to detect `Error` errors to apply the current log. Otherwise, a generic error might be logged instead.

notaphplover (changes requested):
Lets add some logic to detect `Error` errors. The unit test should be updated accordingly.

notaphplover (approved):

notaphplover (approved):

PR #498, opened 2025-03-15
Update InversifyHttpAdapter with readonly logger

notaphplover:
### Changed
- Update InversifyHttpAdapter with readonly logger.

### DIFF
--- a/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
+++ b/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
@@ -26,19 +26,30 @@ export abstract class InversifyHttpAdapter<
 > {
   readonly #container: Container;
   readonly #httpAdapterOptions: InternalHttpAdapterOptions;
-  #logger: Logger;
+  readonly #logger: Logger;
   readonly #routerExplorer: RouterExplorer;
 
   constructor(container: Container, httpAdapterOptions?: HttpAdapterOptions) {
     this.#container = container;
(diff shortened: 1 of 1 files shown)


packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -41,6 +41,17 @@ export abstract class InversifyHttpAdapter<
     await this.#registerControllers();
   }
 
+  #buildLogger(httpAdapterOptions: HttpAdapterOptions | undefined): Logger {
```
Adrianmjim, line 44: Maybe, parameter should be optional? like `parseHttpAdapterOptions`. Or `parseHttpAdapterOptions` should have a posible undefined parameter. idk, I'm just talking about consistency

Adrianmjim (commented):
One question about consistency

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
notaphplover: My current approach is providing optional parameters to public interfaces, but I honeslty don't know how strict I'm about being consistent in this case :sweat_smile:

Adrianmjim: That's a optimal approach for me. I will refactor in future with that standard in mind

Adrianmjim (approved):

PR #499, opened 2025-03-16
Update http adapter with context

notaphplover:
### Changed
- Updated `InversifyHttpAdapter` with non abstract build method.

### DIFF
--- a/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
+++ b/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
@@ -23,6 +23,7 @@ export abstract class InversifyHttpAdapter<
   TRequest,
   TResponse,
   TNextFunction extends (err?: unknown) => void,
+  TContext,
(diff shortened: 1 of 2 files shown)


packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -355,6 +363,7 @@ export abstract class InversifyHttpAdapter<
   ): void;
 
   protected abstract _buildRouter(
+    context: TContext,
     path: string,
     routerParams: RouterParams<TRequest, TResponse, TNextFunction>[],
     guardList: RequestHandler<TRequest, TResponse, TNextFunction>[] | undefined,
```
notaphplover, line 369: 5 params are maybe too many params, does it make sense to rename `RouterParams`to `RouteParams` or `EndpointParams` and extract some `ControllerParams` to cover `path`, `routerParams`, `guardList` and `middlewareList`?

PR #500, opened 2025-03-16
Fix applyMiddleware and useGuard parameter type

Adrianmjim:
### Fixed
- Fix applyMiddleware parameter type
- Fix useGuard parameter type

### DIFF
--- a/packages/http/libraries/core/src/http/decorators/ApplyMiddleware.ts
+++ b/packages/http/libraries/core/src/http/decorators/ApplyMiddleware.ts
@@ -2,13 +2,15 @@ import {
   getReflectMetadata,
   setReflectMetadata,
 } from '@inversifyjs/reflect-metadata-utils';
+import { Newable } from 'inversify';
 
 import { controllerMethodMiddlewareMetadataReflectKey } from '../../reflectMetadata/data/controllerMethodMiddlewareMetadataReflectKey';
 import { controllerMiddlewareMetadataReflectKey } from '../../reflectMetadata/data/controllerMiddlewareMetadataReflectKey';
 import { ControllerFunction } from '../models/ControllerFunction';
+import { Middleware } from '../models/Middleware';
 
 export function applyMiddleware(
-  ...middlewareList: NewableFunction[]
+  ...middlewareList: Newable<Middleware<unknown, unknown, unknown>>[]
 ): ClassDecorator & MethodDecorator {
   return (
     target: object,
--- a/packages/http/libraries/core/src/http/decorators/ApplyMiddleware.ts
+++ b/packages/http/libraries/core/src/http/decorators/ApplyMiddleware.ts
@@ -10,7 +10,7 @@ import { ControllerFunction } from '../models/ControllerFunction';
 import { Middleware } from '../models/Middleware';
 
 export function applyMiddleware(
-  ...middlewareList: Newable<Middleware<unknown, unknown, unknown>>[]
+  ...middlewareList: Newable<Middleware>[]
 ): ClassDecorator & MethodDecorator {
   return (
     target: object,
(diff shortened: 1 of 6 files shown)


packages/http/libraries/core/src/http/decorators/ApplyMiddleware.ts
```diff
@@ -2,13 +2,15 @@ import {
...
+import { Newable } from 'inversify';
 
 import { controllerMethodMiddlewareMetadataReflectKey } from '../../reflectMetadata/data/controllerMethodMiddlewareMetadataReflectKey';
 import { controllerMiddlewareMetadataReflectKey } from '../../reflectMetadata/data/controllerMiddlewareMetadataReflectKey';
 import { ControllerFunction } from '../models/ControllerFunction';
+import { Middleware } from '../models/Middleware';
 
 export function applyMiddleware(
-  ...middlewareList: NewableFunction[]
+  ...middlewareList: Newable<Middleware<unknown, unknown, unknown>>[]
```
notaphplover, line 13:
```suggestion
  ...middlewareList: Newable<Middleware<any, any, any>>[]
```

`TRequest`, `TResponse` and `TNext` are contravariant, so `unknown` is a strong constraint here. Let's usa `any` instead

packages/http/libraries/core/src/http/decorators/UseGuard.spec.ts
```diff
@@ -6,19 +6,21 @@ import {
...
 import { controllerGuardMetadataReflectKey } from '../../reflectMetadata/data/controllerGuardMetadataReflectKey';
 import { controllerMethodGuardMetadataReflectKey } from '../../reflectMetadata/data/controllerMethodGuardMetadataReflectKey';
+import { Guard } from '../guard/Guard';
 import { useGuard } from './UseGuard';
 
 describe(useGuard.name, () => {
   describe('having a ClassDecorator', () => {
     describe('when called and getReflectMetadata returns undefined', () => {
-      let middlewareFixture: NewableFunction;
+      let middlewareFixture: Newable<Guard<unknown>>;
```
notaphplover, line 19: same

packages/http/libraries/core/src/http/decorators/UseGuard.ts
```diff
@@ -2,13 +2,15 @@ import {
...
+import { Newable } from 'inversify';
 
 import { controllerGuardMetadataReflectKey } from '../../reflectMetadata/data/controllerGuardMetadataReflectKey';
 import { controllerMethodGuardMetadataReflectKey } from '../../reflectMetadata/data/controllerMethodGuardMetadataReflectKey';
+import { Guard } from '../guard/Guard';
 import { ControllerFunction } from '../models/ControllerFunction';
 
 export function useGuard(
-  ...guardList: NewableFunction[]
+  ...guardList: Newable<Guard<unknown>>[]
```
notaphplover, line 13: Same

notaphplover (changes requested):
`any` is not the way to go in most cases, but generics targeting function arguments is an exception. To avoid propagating `any` in the source code, I would suggest to update generics default value to be `any`.

For example:

```ts
export interface Guard<TRequest = any> {
  activate(request: TRequest): Promise<boolean> | boolean;
  getHttpResponse?(): HttpResponse;
}
```

packages/http/libraries/core/src/http/decorators/ApplyMiddleware.ts
Adrianmjim: Fixed

adrianmjim:
Nice tip, I did not know about that. Thanks you so much!

notaphplover (approved):

PR #540, opened 2025-03-25
[WIP] [unplanned feature] Add cors support

adrianmjim:
### Added
- Add cors support to HttpAdapterOptions
- Add cors support to InversifyExpressHttpAdapter

### DIFF
--- a/packages/http/libraries/core/package.json
+++ b/packages/http/libraries/core/package.json
@@ -5,13 +5,14 @@
   },
   "description": "InversifyJs http core package",
   "dependencies": {
-    "@inversifyjs/reflect-metadata-utils": "workspace:*",
-    "@inversifyjs/logger": "workspace:*"
+    "@inversifyjs/logger": "workspace:*",
+    "@inversifyjs/reflect-metadata-utils": "workspace:*"
   },
   "devDependencies": {
     "@stryker-mutator/core": "8.7.1",
     "@stryker-mutator/typescript-checker": "8.7.1",
     "@stryker-mutator/vitest-runner": "8.7.1",
+    "@types/cors": "^2.8.17",
     "@types/node": "22.13.13",
     "@vitest/coverage-v8": "3.0.9",
     "eslint": "9.23.0",
--- a/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
+++ b/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
@@ -1,4 +1,5 @@
 import { ConsoleLogger, Logger } from '@inversifyjs/logger';
+import { CorsOptions } from 'cors';
 import { Container } from 'inversify';
 
 import { buildRouterExplorerControllerMetadataList } from '../../routerExplorer/calculations/buildRouterExplorerControllerMetadataList';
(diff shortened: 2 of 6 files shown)


packages/http/libraries/core/package.json
```diff
@@ -5,13 +5,14 @@
...
-    "@inversifyjs/reflect-metadata-utils": "workspace:*",
-    "@inversifyjs/logger": "workspace:*"
+    "@inversifyjs/logger": "workspace:*",
+    "@inversifyjs/reflect-metadata-utils": "workspace:*"
   },
   "devDependencies": {
     "@stryker-mutator/core": "8.7.1",
     "@stryker-mutator/typescript-checker": "8.7.1",
     "@stryker-mutator/vitest-runner": "8.7.1",
+    "@types/cors": "^2.8.17",
```
notaphplover, line 15:
```suggestion
    "@types/cors": "2.8.17",
```

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -53,10 +55,29 @@ export abstract class InversifyHttpAdapter<
...
+    httpAdapterOptions: HttpAdapterOptions | undefined,
+  ): CorsOptions {
+    if (
+      httpAdapterOptions?.cors === undefined ||
+      typeof httpAdapterOptions.cors === 'boolean'
+    ) {
+      return {
+        methods: 'GET,HEAD,PUT,PATCH,POST,DELETE',
+        optionsSuccessStatus: 204,
+        origin: '*',
```
notaphplover, line 68: I don't know if we want to encourage the use of a wildcard origin, it's considered a bad practice

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -53,10 +55,29 @@ export abstract class InversifyHttpAdapter<
...
 
+  #buildCorsOptions(
+    httpAdapterOptions: HttpAdapterOptions | undefined,
+  ): CorsOptions {
+    if (
+      httpAdapterOptions?.cors === undefined ||
+      typeof httpAdapterOptions.cors === 'boolean'
+    ) {
+      return {
+        methods: 'GET,HEAD,PUT,PATCH,POST,DELETE',
```
notaphplover, line 66: Should we add `OPTIONS`? I think some browsers do options call before doing some cors calls, I think allowing OPTIONS would be nice

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -27,15 +28,16 @@ export abstract class InversifyHttpAdapter<
   TResponse,
   TNextFunction extends (err?: unknown) => void,
 > {
+  protected readonly corsOptions: CorsOptions;
```
notaphplover, line 31: this adapter is agnostic, but it's relying on `cors` which is coupled to express. What would be the approach in hono / uwebsockets?

notaphplover (commented):
I added some comments. Let's clarify the implementation details for some missing adapters before merging

adrianmjim:
I think We have to block this feature at the moment. We need to take some arch decisions before

PR #542, opened 2025-03-25
Add stream response support

Adrianmjim:
### Added
- Add stream response support

### DIFF
--- a/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
+++ b/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
@@ -1,5 +1,6 @@
 import { ConsoleLogger, Logger } from '@inversifyjs/logger';
 import { Container } from 'inversify';
+import { Stream } from 'stream';
 
 import { buildRouterExplorerControllerMetadataList } from '../../routerExplorer/calculations/buildRouterExplorerControllerMetadataList';
 import { ControllerMethodParameterMetadata } from '../../routerExplorer/model/ControllerMethodParameterMetadata';
@@ -230,7 +231,8 @@ export abstract class InversifyHttpAdapter<
     value: ControllerResponse,
     statusCode?: HttpStatusCode,
   ): unknown {
-    let body: object | string | number | boolean | undefined = undefined;
+    let body: object | string | number | boolean | Stream | undefined =
+      undefined;
     let httpStatusCode: HttpStatusCode | undefined = statusCode;
 
     if (HttpResponse.is(value)) {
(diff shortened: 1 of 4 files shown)


packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -248,8 +250,10 @@ export abstract class InversifyHttpAdapter<
       return this._replyText(request, response, body);
     } else if (body === undefined || typeof body === 'object') {
       return this._replyJson(request, response, body);
-    } else {
+    } else if (typeof body === 'number' || typeof body === 'boolean') {
       return this._replyText(request, response, JSON.stringify(body));
+    } else {
+      return this._replyStream(request, response, body);
```
notaphplover, line 256: This code is unreachable. A `Stream` `body` is an `object`.

packages/http/libraries/core/src/http/responses/HttpResponse.ts
```diff
@@ -1,3 +1,5 @@
+import Stream from 'stream';
```
notaphplover, line 1:
```suggestion
import Stream from 'node:stream';
```

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -1,5 +1,6 @@
 import { ConsoleLogger, Logger } from '@inversifyjs/logger';
 import { Container } from 'inversify';
+import { Stream } from 'stream';
```
notaphplover, line 3:
```suggestion
import { Stream } from 'node:stream';
```

packages/http/libraries/express/src/adapter/InversifyExpressHttpAdapter.ts
```diff
@@ -14,6 +14,7 @@ import express, {
   Router,
 } from 'express';
 import { Container } from 'inversify';
+import { Stream } from 'stream';
```
notaphplover, line 17:
```suggestion
import { Stream } from 'node:stream';
```

notaphplover (changes requested):
Nice, let's maybe use instance of to detect Stream objects when `typeof body === 'object'.

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
Adrianmjim: Fixed

packages/http/libraries/core/src/http/responses/HttpResponse.ts
Adrianmjim: Fixed

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
Adrianmjim: Fixed

packages/http/libraries/express/src/adapter/InversifyExpressHttpAdapter.ts
Adrianmjim: Fixed

notaphplover (approved):

notaphplover (approved):

PR #543, opened 2025-03-25
Add setHeader decorator

Adrianmjim:
### Added
- Add setHeader decorator
- Add headerMetadata to RouterExplorer
- Add setHeader logic to InversifyHttpAdapter
- Add setHeader logic to InversifyExpressHttpAdapter

### DIFF
--- a/packages/http/libraries/core/src/http/decorators/SetHeader.ts
+++ b/packages/http/libraries/core/src/http/decorators/SetHeader.ts
@@ -1,0 +1,29 @@
+import {
+  getReflectMetadata,
+  setReflectMetadata,
+} from '@inversifyjs/reflect-metadata-utils';
+
+import { controllerMethodHeaderMetadataReflectKey } from '../../reflectMetadata/data/controllerMethodHeaderMetadataReflectKey';
+import { ControllerFunction } from '../models/ControllerFunction';
+
+export function setHeader(key: string, value: string): MethodDecorator {
+  return (
+    _target: object,
+    _key: string | symbol,
+    descriptor: PropertyDescriptor,
+  ): void => {
+    const headerMetadata: Map<string, string> =
+      getReflectMetadata(
+        descriptor.value as ControllerFunction,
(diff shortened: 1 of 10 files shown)


packages/http/libraries/core/src/http/decorators/SetHeader.ts
```diff
@@ -0,0 +1,29 @@
...
+import { controllerMethodHeaderMetadataReflectKey } from '../../reflectMetadata/data/controllerMethodHeaderMetadataReflectKey';
+import { ControllerFunction } from '../models/ControllerFunction';
+
+export function setHeader(key: string, value: string): MethodDecorator {
+  return (
+    _target: object,
+    _key: string | symbol,
+    descriptor: PropertyDescriptor,
+  ): void => {
+    const headerMetadata: Map<string, string> =
```
notaphplover, line 15: Do we want to support multiple values in order to combine them? Refer to https://www.rfc-editor.org/rfc/rfc9110.html#name-field-lines-and-combined-fi for more information.

For example, developers might want to `@setHeader('Cache-Control', 'no-cache') @setHeader('Cache-Control', 'no-store')`

packages/http/libraries/core/src/http/decorators/SetHeader.ts
```diff
@@ -0,0 +1,29 @@
...
+      getReflectMetadata(
+        descriptor.value as ControllerFunction,
+        controllerMethodHeaderMetadataReflectKey,
+      ) ?? new Map<string, string>();
+
+    headerMetadata.set(key, value);
```
notaphplover, line 21: headers are not case sensitive, consider https://github.com/fastify/help/issues/71
```suggestion
    headerMetadata.set(key.toLowerCase(), value);
```

Adrianmjim: Fixed

notaphplover (approved):

PR #563, opened 2025-03-30
Polish http benchmarks package

notaphplover:
### Added
- Added http benchmarks readme file.

### Changed
- Renamed setup modules to `setup.ts`.
- Updated `knip` http benchmarks config with missing entrypoints.


knip.ts
```diff
@@ -89,26 +84,14 @@ export default {
...
-      ],
-      ignore: [
-        "src/constant/*.ts",
-        "src/scenario/*/setUp*.ts",
+      entry: [
+        "src/bin/run.ts",
         "src/k6/scenario/*.ts",
-        "src/scenario/nestJS/**/*.ts",
+        "src/scenario/*/setup.ts",
+      ],
```
notaphplover, line 91: @adrianmjim this way we tell knip which entrypoints should be analyzed. If we don't provide them, knip is likelly to fail detecting unused content

adrianmjim (approved):

PR #577, opened 2025-04-02
Replace Stream with ReadableStream

adrianmjim:
### Refactored
- Replace Stream with ReadableStream in ControllerResponse, InversifyHttpAdapter and InversifyExpressHttpAdapter


notaphplover (changes requested):
As copilot says, `Readable.fromWeb` expects a Web ReadableStream rather than a Node.js stream

PR #578, opened 2025-04-02
Update container with load and unload sync

notaphplover:
### Changed
- Updated `Container` with `loadSync`.
- Updated `Container` with `unloadSync`.


BurningEnlightenment:
Thank you for implementing this. Async container loading is a real headache if you need to build a service graph before some runtime event like electron's [app#ready] event fires[^1].

[^1]: some of electron's configuration methods can only be called before app#ready fires

[app#ready]: https://www.electronjs.org/docs/latest/api/app#event-ready

PR #581, opened 2025-04-02
Add body decorator e2e tests

adrianmjim:


packages/http/tools/e2e-tests/src/warrior/step-definitions/thenDefinitions.ts
```diff
@@ -17,31 +31,41 @@ async function thenResponseContainsTheCorrectUrlParametersByName(
...
   const response: Response =
     getServerResponseOrFail.bind(this)(parsedResponseAlias);
-  // eslint-disable-next-line @typescript-eslint/no-magic-numbers
-  assert.ok(response.status >= 200 && response.status < 300);
+
+  const warriorCreationResponse: WarriorCreationResponse =
+    (await response.json()) as WarriorCreationResponse;
+
+  assert(warriorCreationResponse.name === 'Samurai');
+  assert(warriorCreationResponse.type === 'Melee');
```
notaphplover, line 49: Let's use enums instead of magic strings

packages/http/tools/e2e-tests/src/warrior/step-definitions/thenDefinitions.ts
```diff
@@ -17,31 +31,41 @@ async function thenResponseContainsTheCorrectUrlParametersByName(
 
   const warriorWithId: WarriorWithId = (await response.json()) as WarriorWithId;
 
-  assert.ok(warriorWithId.id === '123');
+  assert(warriorWithId.id === '123');
```
notaphplover, line 34: We should compare it the request param. We can probably use the world to accomplish this

notaphplover (changes requested):
I added some minor changes. Overall looks good

packages/http/tools/e2e-tests/src/warrior/step-definitions/thenDefinitions.ts
adrianmjim: I think this is not related to this PR. I will submit a fix in another PR

packages/http/tools/e2e-tests/src/warrior/step-definitions/thenDefinitions.ts
adrianmjim: Sure, fixed

notaphplover (approved):

PR #594, opened 2025-04-04
Add next decorator and related handler logic

adrianmjim:
### Added
- Add next decorator
- Add useNativeHandler metadata
### Refactored
- Refactor InversifyHttpAdapter to implement useNativeHandler logic


packages/http/libraries/core/src/http/models/MiddlewareHandler.ts
```diff
@@ -1,10 +0,0 @@
-export type MiddlewareHandler<
-  TRequest,
-  TResponse,
-  TNextFunction,
-  TResult = void,
-> = (
-  req: TRequest,
-  res: TResponse,
-  next: TNextFunction,
-) => Promise<TResult> | TResult;
```
notaphplover, line 10: I would keep the type. They look the same but they are not the same.

notaphplover (approved):
It looks good. I would keep MiddlewareHandler though

packages/http/libraries/core/src/http/models/MiddlewareHandler.ts
adrianmjim: Sure, don't merge, I will readd it

notaphplover (approved):

PR #607, opened 2025-04-06
Add custom decorator support

adrianmjim:
### Added
- Add custom decorator


packages/http/libraries/core/src/http/calculations/createCustomParameterDecorator.ts
```diff
@@ -0,0 +1,64 @@
...
+import { controllerMethodParameterMetadataReflectKey } from '../../reflectMetadata/data/controllerMethodParameterMetadataReflectKey';
+import { ControllerMethodParameterMetadata } from '../../routerExplorer/model/ControllerMethodParameterMetadata';
+import { Controller } from '../models/Controller';
+import { ControllerFunction } from '../models/ControllerFunction';
+import { CustomParameterDecoratorHandler } from '../models/CustomParameterDecoratorHandler';
+import { RequestMethodParameterType } from '../models/RequestMethodParameterType';
+import { insertParameterMetadata } from './insertParameterMetadata';
+
+export function createCustomParameterDecorator<TRequest, TResponse, TResult>(
+  handler: CustomParameterDecoratorHandler<TRequest, TResponse, TResult>,
```
notaphplover, line 17:
```suggestion
  handler: CustomParameterDecoratorHandler<TRequest, TResponse, TResult>,
```
This handler cannot have dependencies like middlewares, wards and controllers, right? This is an important limitation.

notaphplover: I don't think we want. Middleware feels a superior version of this approach. Now I realize that's probably the motivation for executing middlewares before guards, to implement an auth provider as a middleware so the user guard is aware of the user

adrianmjim: I don't think middlewares should go first, guards are usually responsible for that task. Anyway, this functionality is not designed for authentication, although it can be used in that form.

notaphplover: > I don't think middlewares should go first, guards are usually responsible for that task

We can go that way. I know we took this decission some months ago, but I've been thinking there might be middlewares that are supposed to run before any guard, such helmet. But I don't see a compeling reason to not to run a guard after a middleware.

> Anyway, this functionality is not designed for authentication, although it can be used in that form.
I'm aware, I simply believe the middleware gives you the same than a custom decorator and it allows you defining dependencies in the middleware. What would be the advantages of this feature?

adrianmjim: > We can go that way. I know we took this decission some months ago, but I've been thinking there might be middlewares that are supposed to run before any guard, such as helmet. But I don't see a compeling reason to not to run a guard after a middleware.
>
Well, in that case we can reorder that logic easily. But a middleware is not supposed to authenticate a request, that's a guard responsibility. If You want to implement a midleware to authenticate request and then check user information, you can implement a double guard. Each guard with their own responsibility, obsiously.
> I'm aware, I simply believe middlewares gives you the same than a custom decorator and it allows you defining dependencies in the middleware. What would be the advantages of this feature?
>
I think I created a confusion here, sorry. Let me explain some concept: Middleware and guards can modify the request to assign dependencies but that's not the feature proposed. Here, we do not attempt to assign dependencies to the request or response. Instead, we provide a simple functionality for the user to implement their own parameter decorators. Any assignment phase should be performed prior to the execution of the handler.

notaphplover (approved):
LGTM!

PR #627, opened 2025-04-12
Extract e2e request parameter

notaphplover:
### Added
- Added `RequestParameter`.

### Changed
- Updated world requests to be a map of `RequestParameter`.
- Updated given and then scenarios to properly handle request parameters.

### Motivation
Request info is often hard to get at assertion time. Url params must be extracted from the url, request bodies are flushed and no longer available... it's not easy at all. `RequestParameter` allow us to accomodate request relevant data in a friendly way so we can use it at assertion time.


adrianmjim (approved):
:heart:

Issue #522, opened 2025-03-20
[Feature]:  Implement validation support


Adrianmjim:
### 🚀 Feature Proposal

Implement http validation with suggested packages:
- [zod](https://zod.dev)
- [ajv](https://ajv.js.org)
- [class-validator](https://github.com/typestack/class-validator)

### Motivation

Implementing http validation is a must in order to provide usable http tools.

### Pitch

This feature is related to http-core package

acommodari:
standard schema could be useful here
https://standardschema.dev/

notaphplover:
Hey @acommodari, sry for the late reply.

> standard schema could be useful here

I see. It can be super interesting. We could have a standard schema integration which would kill many birds with one stone. We'll go for it in further releases for sure, but unless someone wants to submit a PR I prefer to not to go for it in the first release to speed up the process (you've wait for long enough).

**Edit**: At the end I couldn't resist: #1018.

notaphplover:
Every requested validator is now supported 🎉

PR #646, opened 2025-04-16
Add Pipe support

adrianmjim:
### Added
- Add Pipe contract interface
- Add pipe support for default decorators
- Add pipe support for custom decorators

### Refactored
- Refactor requestParam decorator handling
- Move functions to correct folders

### Motivation
This is a initial phase of #522


packages/http/libraries/core/src/http/typeguard/isPipe.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+        valueFixture = undefined;
+
+        result = isPipe(valueFixture);
+      });
+
+      it('should return false', () => {
+        expect(result).toBe(false);
+      });
+    });
+  });
```
notaphplover, line 41: Let's have different cases to ensure we cover all the flows:

1. undefined.
2. null.
3. Object with no `execute` property.
4. Object with non function `execute` property.

I love using describe.each in these kind of tests :slightly_smiling_face:

packages/http/libraries/core/src/http/typeguard/isApplyMiddlewareOptions.ts
```diff
@@ -7,6 +7,8 @@ export function isApplyMiddlewareOptions(
     value as ApplyMiddlewareOptions;
 
   return (
+    value !== undefined &&
```
notaphplover, line 10:
```suggestion
    value != undefined &&
```

packages/http/libraries/core/src/http/typeguard/isApplyMiddlewareOptions.ts
```diff
@@ -7,6 +7,8 @@ export function isApplyMiddlewareOptions(
     value as ApplyMiddlewareOptions;
```
notaphplover, line 7: Let's use the Partial here as well (see below comments at `isPipe`)

packages/http/libraries/core/src/http/calculations/requestParamFactory.ts
```diff
@@ -0,0 +1,38 @@
+import { Newable } from 'inversify';
+
+import { ControllerMethodParameterMetadata } from '../../routerExplorer/model/ControllerMethodParameterMetadata';
+import { CustomParameterDecoratorHandler } from '../models/CustomParameterDecoratorHandler';
+import { RequestMethodParameterType } from '../models/RequestMethodParameterType';
+import { Pipe } from '../pipe/model/Pipe';
+import { requestParam } from './requestParam';
+
+export function requestParamFactory(
```
notaphplover, line 9:
```suggestion
export function buildRequestParameterDecorator(
```
What do you think about this alternative name?

packages/http/libraries/core/src/http/calculations/requestParamFactory.ts
```diff
@@ -0,0 +1,38 @@
...
+  parameterType: RequestMethodParameterType,
+  parameterPipeList: (Newable<Pipe> | Pipe)[],
+  parameterNameOrPipe?: string | (Newable<Pipe> | Pipe),
+  customParameterDecoratorHandler?: CustomParameterDecoratorHandler,
```
notaphplover, line 13:
```suggestion
  parameterName: string | undefined,
  customParameterDecoratorHandler: CustomParameterDecoratorHandler | undefined
```
Let's simplify the method signature

notaphplover (changes requested):
I need to have it a closed look since this PR is big, but I added some minor changes to apply

packages/http/libraries/core/src/http/typeguard/isPipe.ts
```diff
@@ -0,0 +1,11 @@
+import { Pipe } from '../pipe/model/Pipe';
+
+export function isPipe(value: unknown): value is Pipe {
+  const pipe: Pipe = value as Pipe;
+
+  return (
+    value !== undefined &&
```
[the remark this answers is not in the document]
adrianmjim: I prefer a more explicit approach

packages/http/libraries/core/src/http/typeguard/isPipe.spec.ts
adrianmjim: Fixed

packages/http/libraries/core/src/http/typeguard/isApplyMiddlewareOptions.ts
adrianmjim: Fixed with a explicit approach

packages/http/libraries/core/src/http/typeguard/isApplyMiddlewareOptions.ts
adrianmjim: Fixed

packages/http/libraries/core/src/http/calculations/requestParamFactory.ts
adrianmjim: This way we remove some funcionality. This function is supposed to abstract this complexity

notaphplover: I think this method signature makes the code harder to read. I understand this signature in decorators such as `body`, but I would prefer to have simpler arguments here: one parameterName and a list of pipes. The duplicated logic can be avoided implementing some utils function. If you want I can go for it

packages/http/libraries/core/src/error/models/InversifyHttpAdapterErrorKind.ts
```diff
@@ -1,4 +1,5 @@
 export enum InversifyHttpAdapterErrorKind {
   noControllerFound,
   requestParamIncorrectUse,
+  pipeError,
```
notaphplover, line 4: Let's sort them alphabetically

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -299,12 +320,23 @@ export abstract class InversifyHttpAdapter<
...
+            if (
+              this.#globalPipeList.length +
+                controllerMethodParameterMetadata.pipeList.length >
+              0
+            ) {
+              return this.#applyPipeList(params, index, [
+                ...this.#globalPipeList,
+                ...controllerMethodParameterMetadata.pipeList,
+              ]);
+            }
```
notaphplover, line 339:
```suggestion
            return this.#applyPipeList(params, index, [
              ...this.#globalPipeList,
              ...controllerMethodParameterMetadata.pipeList,
            ]);
```

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -49,6 +55,11 @@ export abstract class InversifyHttpAdapter<
     this.#container = container;
     this.#logger = this.#buildLogger(httpAdapterOptions);
     this.httpAdapterOptions = this.#parseHttpAdapterOptions(httpAdapterOptions);
+    this.#globalPipeList = [];
```
notaphplover, line 58: Just to be clear: this pipes are applied to every parameter of every handler, right?

packages/http/libraries/core/src/http/pipe/model/Pipe.ts
```diff
@@ -0,0 +1,3 @@
+export interface Pipe<TInput = unknown, TOutput = unknown> {
+  execute(input: TInput): Promise<TOutput> | TOutput;
```
notaphplover, line 2: I think we are missing some metadata, right? It makes sense to me to give the pipe some param metadata such as the target class, the target method and the parameter index. Pipes might require this medatada to access related reflected metadata to perform whatever logic they want.

packages/http/libraries/core/src/http/calculations/requestParamFactory.ts
adrianmjim: I understand your point, but this function has been created with the purpose of being that util function which has the responsibility to avoid such complexity; it is not a matter of avoiding certain logic. If the contract is configured in the proposed way, the decorators will always have to use a parameterName to be able to use pipes, and that is not the desired behavior.

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
adrianmjim: Yeah, that's the idea

packages/http/libraries/core/src/error/models/InversifyHttpAdapterErrorKind.ts
adrianmjim: Fixed

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
adrianmjim: Fixed

packages/http/libraries/core/src/http/pipe/model/Pipe.ts
adrianmjim: I understand having a target class, but I don't find reasons to have the rest of the metadata. What kind of approach do you propose for the pipes that requires such metadata?

notaphplover: Validation pipes probably require the both the method name and the parameter index in order to read related reflact metadata, so I think it might be useful.

Regarding the approach, we can update `execute` with a `metadata: PipeMetadata` param:

```ts
interface PipeMetadata {
  methodName: string | symbol;
  paramIndex: number;
  paramType: RequestMethodParameterType;
  type: Newable,
}
```

What do you think about it?

packages/http/libraries/core/src/http/calculations/requestParamFactory.ts
notaphplover: > I understand your point, but this function has been created with the purpose of being that util function which has the responsibility to avoid such complexity; it is not a matter of avoiding certain logic

I see. Not my cup of tea, but let's keep it that way then :slightly_smiling_face:.

> decorators will always have to use a parameterName to be able to use pipes, and that is not the desired behavior

Not really, parameterName will be `undefined` in that case.

packages/http/libraries/core/src/http/pipe/model/Pipe.ts
notaphplover: To ellaborate on this, let's asume we want to implement a global validation Pipe:

```ts
export class ZodValidationPipe implements Pipe {
  // Super super implementation
}
```

Not, let's say we want to design the validation library in a way users can provide validation schemas as the controller param decorators:

```ts
import { body, controller, POST } from '@inversifyjs/http-core';
import { zodValidationSchema } from '@inversifyjs/validation-zod';

@controller('/foo')
export class MyController {
  @POST()
  public async createFoo(
    @body()
    @zodValidationSchema(fooZodSchema)
    body: Foo,
  ): Promise<Foo> {
    return foo;
  }
}

```

How is `ZodValidationPipe` supposed to access to `zodValidationSchema` metadata if we don't give it the metadata I proposed?

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -212,12 +226,24 @@ export abstract class InversifyHttpAdapter<
...
         this.#printError(error);
-        return this.#reply(req, res, new InternalServerErrorHttpResponse());
+
+        if (
+          InversifyHttpAdapterError.isErrorOfKind(
+            error,
+            InversifyHttpAdapterErrorKind.pipeError,
+          )
+        ) {
+          return this.#reply(req, res, new BadRequestHttpResponse());
```
notaphplover, line 236: The user might want to fail with a different http error code. 401, 403, 404, 409, 422 might be possible codes as well in some use cases.

We would need a way to allow the user to throw a custom error code.

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -299,12 +325,27 @@ export abstract class InversifyHttpAdapter<
...
+              [
+                ...this.#globalPipeList,
+                ...controllerMethodParameterMetadata.pipeList,
+              ],
+              {
+                methodName: controllerMethodKey,
+                parameterIndex: index,
+                parameterMethodType:
+                  controllerMethodParameterMetadata.parameterType,
+                targetClass: controller,
```
notaphplover, line 346: You're passing the controller, but we must pass the controller class. There's a `getBaseType` that can be used for the same purpose

notaphplover (changes requested):
The code looks good, but we need to polish some details before merging

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
adrianmjim: I added a simple approach

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
adrianmjim: Ups, mb. Fixed

packages/http/libraries/core/src/http/pipe/model/Pipe.ts
```diff
@@ -0,0 +1,7 @@
+import { HttpResponse } from '../../responses/HttpResponse';
+import { PipeMetadata } from './PipeMetadata';
+
+export interface Pipe<TInput = unknown, TOutput = unknown> {
+  execute(input: TInput, metadata: PipeMetadata): Promise<TOutput> | TOutput;
+  getHttpResponse?(): HttpResponse;
```
notaphplover, line 6: I see we're applying the same pattern to the guard.

What if we want to set different status codes / messages depending on the received input?

We can maybe merge now and think how to solve this use case for guards / pipes in another PR

notaphplover (approved):

PR #654, opened 2025-04-18
Fix fastify reponse

adrianmjim:
### Motivation
Fixes #631

### DIFF
--- a/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
+++ b/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
@@ -205,67 +205,79 @@ export abstract class InversifyHttpAdapter<
     response: TResponse,
     next: TNextFunction,
   ): Promise<unknown[]> {
-    return Promise.all(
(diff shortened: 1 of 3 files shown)


packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -205,67 +205,79 @@ export abstract class InversifyHttpAdapter<
...
+            break;
           }
-        },
-      ),
-    );
+        }
+      }
+    }
+
+    return handlerParams;
```
notaphplover, line 280: This code changes are done so fastify can integration can be fixed, but then it's coupled to fastify, for awaiting the body could be harmful in other integration.

I understand awaiting some promise like objects like the fastify `Response` is not what we want and might have undesired side effects. Specific adapters are the ones who know this, not the abstraction.

Does it make sense to you? If so, I have a couple of ideas of how to implement this in a way we depend on an specific adapter settings

notaphplover:
I submitted #661 with my proposal

PR #661, opened 2025-04-20
Update http adapter with awaitable param types config

notaphplover:
### Changed
- Updated http core to expose `RequestMethodParameterType`.
- Updated `InversifyHttpAdapter._getBody` to not to be forced to return a promise value.
- Updated copilot docs with `vitest` usage.
- Updated turbo `e2e` task with missing dependency.
- Updated `InversifyHttpAdapter` to properly handle request method params.


packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -32,10 +32,20 @@ export abstract class InversifyHttpAdapter<
...
   protected readonly httpAdapterOptions: InternalHttpAdapterOptions;
+  readonly #awaitableRequestMethodParamTypes: Set<RequestMethodParameterType>;
   readonly #container: Container;
   readonly #logger: Logger;
 
-  constructor(container: Container, httpAdapterOptions?: HttpAdapterOptions) {
+  constructor(
+    container: Container,
+    httpAdapterOptions: HttpAdapterOptions | undefined,
+    awaitableRequestMethodParamTypes:
```
adrianmjim, line 42: Maybe, could this parameter be optional? So we could avoid undefined at some child contructors(majority in this case)

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -205,67 +226,91 @@ export abstract class InversifyHttpAdapter<
     response: TResponse,
     next: TNextFunction,
   ): Promise<unknown[]> {
-    return Promise.all(
+    const params: unknown[] = new Array(
+      controllerMethodParameterMetadataList.length,
+    );
+
+    await Promise.all(
```
adrianmjim, line 233: I think a for loop approach is more explicity for this type of operation. What do You think?

adrianmjim (changes requested):
Some little questions

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
notaphplover: Yeah, but on the other hand Promise.all adds less overhead

adrianmjim (approved):

PR #674, opened 2025-04-26
Add middleware e2e tests

adrianmjim:


packages/http/tools/e2e-tests/src/warrior/middleware/middlewares/UnsuccessfulMiddleware.ts
```diff
@@ -0,0 +1,28 @@
...
+export class UnsuccessfulMiddleware
+  implements Middleware<unknown, unknown, unknown, unknown>
+{
+  public async execute(
+    _request: unknown,
+    response: {
+      send?: () => unknown;
+      text?: () => unknown;
+      status: (code: number) => void;
+    },
```
notaphplover, line 14: This implementation might cause trouble with new adapters. I think we need a class per integration or a way to detect which adapter we are using and use the right specific adapter types

notaphplover (commented):
The code looks good, but we can add some extra assertions.

I'd love the successful middlerare to set a response header so we can verify in the response the header is set and therefore the middleware was called.

I think the unsuccessful case would be even better if we apply a successful middleware before the unsuccessful one, this way we can detect in the response there's the reponse header and, therefore, middlewares are executed in the right order (if the unsuccessful middleware was executed before, it would send the response so the header wouldn't be there)

notaphplover (approved):

notaphplover (approved):

PR #689, opened 2025-05-02
Refactor http build phase

adrianmjim:


packages/http/libraries/core/src/http/calculations/requestParam.ts
```diff
@@ -57,9 +51,10 @@ export function requestParam(
...
         RequestMethodParameterType.RESPONSE
     ) {
-      setReflectMetadata(
-        controllerFunction,
+      setReflectMetadataWithProperty(
+        target,
         controllerMethodUseNativeHandlerMetadataReflectKey,
+        key,
         true,
       );
```
notaphplover, line 59: Let's overload `setReflectMetadata` instead

packages/http/libraries/core/src/http/decorators/UseGuard.ts
```diff
@@ -37,16 +35,17 @@ export function useGuard(
...
       );
     } else {
-      setReflectMetadata(
-        descriptor.value as ControllerFunction,
+      setReflectMetadataWithProperty(
+        target,
         controllerMethodGuardMetadataReflectKey,
+        key,
         guardMetadataList,
       );
```
notaphplover, line 50: Let's rely on updateOwnReflectMetadata instead of getting and setting metadata.

packages/http/libraries/core/src/routerExplorer/calculations/buildRouterExplorerControllerMethodMetadata.ts
```diff
@@ -13,38 +12,58 @@ import { exploreControllerMethodParameterMetadataList } from './exploreControlle
...
 import { exploreControllerMethodUseNativeHandlerMetadata } from './exploreControllerMethodUseNativeHandlerMetadata';
 
-export function buildRouterExplorerControllerMethodMetadata(
-  controller: Controller,
+export function buildRouterExplorerControllerMethodMetadata<
+  TRequest,
+  TResponse,
+  TResult,
+>(
+  controller: NewableFunction,
```
notaphplover, line 20:
```suggestion
  controllerConstructor: NewableFunction,
```

packages/http/libraries/core/src/routerExplorer/model/ControllerMethodParameterMetadata.ts
```diff
@@ -1,13 +1,14 @@
...
 import { RequestMethodParameterType } from '../../http/models/RequestMethodParameterType';
 import { Pipe } from '../../http/pipe/model/Pipe';
 
 export interface ControllerMethodParameterMetadata<
-  TRequest = unknown,
-  TResponse = unknown,
-  TResult = unknown,
+  TRequest = any,
+  TResponse = any,
+  TResult = any,
```
notaphplover, line 11: Why do we need those castings now?

notaphplover (changes requested):
After a first review I added several comments:

1. Let's use get own metadata calls instead of get metadata ones.
2. Let's avoid adding more metadata utils if overloading is good enough.
3. Let's try to rely on the update own metadata util when possible.

In addition, can you add a description with the most notable changes in the PR? It's sort of big so it would help to understand it better :slightly_smiling_face:

PR #713, opened 2025-05-20
Add Container#dispose method to clean up memory from child containers

leoptizmo:
The `dispose` method removes `PlanResultCacheService` subscriptions on the parent container to avoid leaking memory when constructing temporary child containers. It should be called once the child container is no longer in use.

I've converted the `#subscribers` array in `PlanResultCacheService` into a set so that removing child containers' cache services is O(1) rather than O(N) where N is the length of the array.

Fixes inversify/InversifyJS#1810


leoptizmo:
Messed up the branches, closing and reopening on a new branch.

Continued in #715

PR #715, opened 2025-05-21
feat: add Container#dispose method to clean up memory from child containers

leoptizmo:
The `dispose` method removes `PlanResultCacheService` subscriptions on the parent container to avoid leaking memory when constructing temporary child containers. It should be called once the child container is no longer in use.

I've converted the `#subscribers` array in `PlanResultCacheService` into a set so that removing child containers' cache services is O(1) rather than O(N) where N is the length of the array.

Fixes inversify/InversifyJS#1810

### DIFF
--- a/packages/container/libraries/core/src/planning/services/PlanResultCacheService.ts
+++ b/packages/container/libraries/core/src/planning/services/PlanResultCacheService.ts
@@ -54,7 +54,7 @@ export class PlanResultCacheService {
     Map<MetadataName, Map<MetadataTag, Map<unknown, PlanResult>>>
   >[];
 
-  readonly #subscribers: PlanResultCacheService[];
+  readonly #subscribers: Set<PlanResultCacheService>;
 
   constructor() {
     this.#serviceIdToValuePlanMap = this.#buildInitializedMapArray();
@@ -62,7 +62,7 @@ export class PlanResultCacheService {
     this.#namedTaggedServiceIdToValuePlanMap = this.#buildInitializedMapArray();
     this.#taggedServiceIdToValuePlanMap = this.#buildInitializedMapArray();
 
-    this.#subscribers = [];
+    this.#subscribers = new Set();
   }
 
   public clearCache(): void {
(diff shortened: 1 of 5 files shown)


packages/container/libraries/core/src/planning/services/PlanResultCacheService.ts
```diff
@@ -54,15 +54,15 @@ export class PlanResultCacheService {
     Map<MetadataName, Map<MetadataTag, Map<unknown, PlanResult>>>
   >[];
 
-  readonly #subscribers: PlanResultCacheService[];
+  readonly #subscribers: Set<PlanResultCacheService>;
```
notaphplover, line 57: As said, let's try to rely on a `WeakSet` instead

packages/container/libraries/core/src/planning/services/PlanResultCacheService.ts
```diff
@@ -156,7 +156,11 @@ export class PlanResultCacheService {
   }
 
   public subscribe(subscriber: PlanResultCacheService): void {
-    this.#subscribers.push(subscriber);
+    this.#subscribers.add(subscriber);
+  }
+
+  public unsubscribe(subscriber: PlanResultCacheService): void {
+    this.#subscribers.delete(subscriber);
   }
```
notaphplover, line 164: This method should not be required thanks to the `WeakSet` approach

notaphplover (changes requested):
Thank you so much for your contribution! Let's try the `WeakSet` approach before for its simplicity

packages/container/libraries/core/src/planning/services/PlanResultCacheService.ts
leoptizmo: I don't think we can use `WeakSet` because we need to be able to iterate through the elements of the set in the `clearCache` method: https://github.com/inversify/monorepo/blob/fce7fe2023cb7d68c68637948614a2a7e1cc36da/packages/container/libraries/core/src/planning/services/PlanResultCacheService.ts#L73

notaphplover: Hey @leoptizmo, I see, then an array of [WeakRef](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) should do the trick. If we don't need to unsubscribe, an array is good enough

leoptizmo: Thanks for the review and suggestion! I've implemented this suggestion in a new PR as it has some caveats: https://github.com/inversify/monorepo/pull/718

notaphplover:
Clonsing since the fix was accomplished in #718

PR #718, opened 2025-05-23
fix(core): fix child container memory leak (with caveat)

leoptizmo:
As suggested by https://github.com/inversify/monorepo/pull/715#discussion_r2102220927, this PR resolves memory leaks from child containers by storing child cache service subscriptions to the parent cache service using `WeakRef`.

This should fix memory leaks in real world applications which most likely involve [different jobs running in the JS event loop](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#job) between child container instantiations. However, it is still possible for the process to run out of memory in the pathological case where there are no other jobs or not enough jobs between child container instantiations. This is because `WeakRef` imposes a constraint on its target, [preventing it from being garbage collected until a subsequent turn of the event loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef#notes_on_weakrefs).

Fixes https://github.com/inversify/InversifyJS/issues/1810

## Test script template

The below script contains comments of the form `// await job() // (N)` to indicate positions where we can force a job into the event loop and test how that affects memory usage.

This script should be run with `node --expose-gc` to allow it to force garbage collection at specific points.

```ts
const { Container } = require('inversify')
const { setTimeout } = require('node:timers/promises')

const logMemoryUsage = msg => {
  gc()
  console.log(msg, Math.round(process.memoryUsage().heapUsed / 1e6), 'MB')
}

const job = () => setTimeout(0)

;(async () => {
  // New scope so we can log memory usage after GC'ing the container.
  await (async () => {
    const container = new Container()
    const leak = async () => {
      const child = new Container({ parent: container })
      // await job() // (1)
    }

    logMemoryUsage('start:')
    for (let i = 0; i < 100_000; ++i) {
      await leak()
    }
    // await job() // (2)
    logMemoryUsage('after loop GC:')
  })()

  // await job() // (3)
  logMemoryUsage('after container GC:')
})()
```

## Results

For comparison, the last known good version (inversify@7.0.0-alpha.1) has no memory leaks:

| Start | After loop GC | After container GC |
| ----- | ------------- | ------------------ |
| 4 MB  | 4 MB          | 4 MB               |

The current latest version (inversify@7.5.1) has the same leak regardless of where we insert `await job()`:

| Start | After loop GC | After container GC |
| ----- | ------------- | ------------------ |
| 4 MB  | 342 MB        | 4 MB               |

The results for this PR are tabulated below:

| `await job()` position  | Start | After loop GC | After container GC |
| ----------------------- | ----- | ------------- | ------------------ |
| Within loop (1)         | 4 MB  | 8 MB          | 4 MB               |
| Before loop GC (2)      | 4 MB  | 9 MB          | 4 MB               |
| Before container GC (3) | 4 MB  | 348 MB        | 4 MB               |
| N/A                     | 4 MB  | 348 MB        | 344 MB             |

Despite the similarity between cases (1) and (2), all cases but (1) may lead to out-of-memory errors; running case (2) with `--max-old-space-size=200` causes it to crash because the child `PlanResultCacheService` instances cannot be garbage collected until `await job() // (2)` runs, meaning that the loop would require 348 MB of memory like the other cases. (This also explains the discrepancy in the "after loop GC" values between cases (1) and (2), as only case (1) resizes the `#subscribers` array.)

The case with no extra jobs (N/A) is especially interesting to compare against the current latest version. It performs _worse_ due to the constraints imposed by `WeakRef`. To reiterate, this is unlikely to cause problems in the real world, since useful applications tend to require additional jobs in the event loop (I assume).

With these caveats in mind, I'd like to hear the maintainers' thoughts on whether this is an acceptable solution to the memory leak problem.

### DIFF
--- a/.changeset/eager-drinks-behave.md
+++ b/.changeset/eager-drinks-behave.md
@@ -1,0 +1,7 @@
+---
+"@inversifyjs/core": patch
+---
+
+Fixed child container memory leak in most cases
+
+Provided that there is a [job](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#job) in the event loop after constructing a child container, its cache service will eventually be garbage collected along with the child container itself, rather than persisting until the parent container's garbage collection.
--- a/packages/container/libraries/core/src/planning/services/PlanResultCacheService.ts
+++ b/packages/container/libraries/core/src/planning/services/PlanResultCacheService.ts
@@ -26,6 +26,16 @@ export interface GetPlanOptions {
   tag: GetPlanOptionsTagConstraint | undefined;
 }
 
+const MIN_REFS_FOR_REALLOCATION: number = 8;
+const MIN_DEAD_REFS_FOR_REALLOCATION_PERCENTAGE: number = 0.5;
+function shouldReallocate(liveRefCount: number, refCount: number): boolean {
+  return (
+    refCount > MIN_REFS_FOR_REALLOCATION &&
+    refCount - liveRefCount >
+      refCount * MIN_DEAD_REFS_FOR_REALLOCATION_PERCENTAGE
+  );
+}
+
 /**
  * Service to cache plans.
  *
@@ -54,7 +64,7 @@ export class PlanResultCacheService {
     Map<MetadataName, Map<MetadataTag, Map<unknown, PlanResult>>>
   >[];
 
-  readonly #subscribers: PlanResultCacheService[];
+  #subscribers: WeakRef<PlanResultCacheService>[];
 
   constructor() {
     this.#serviceIdToValuePlanMap = this.#buildInitializedMapArray();
@@ -70,8 +80,18 @@ export class PlanResultCacheService {
       map.clear();
     }
 
-    for (const subscriber of this.#subscribers) {
-      subscriber.clearCache();
+    let liveRefCount: number = 0;
+    for (const subscriberRef of this.#subscribers) {
+      const subscriber: PlanResultCacheService | undefined =
+        subscriberRef.deref();
+      if (subscriber) {
+        subscriber.clearCache();
+        ++liveRefCount;
+      }
+    }
+
+    if (shouldReallocate(liveRefCount, this.#subscribers.length)) {
+      this.#compactSubscribersArray(liveRefCount);
     }
   }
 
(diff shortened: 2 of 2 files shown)


notaphplover (commented):
Amazing reasearch, amazing results!

Thank you so much for your work.

Results are great, but I think we can add a final touch.

There's still an small memory leak. If I'm right, the array is not reallocated in your case since no parent binding updates are triggering the `clearCache` method. And it makes perfect sense, that might very well be an usual use case in a long lived application.

I think I missed pointing to another method to update in order to reallocate a subscribers array: `subscribe`.

Subscribe is more delicate for a reason: any realocation strategy turns `subscribe` from `O(1)` to `O(n)`, being `n` the lenght of the subscribers array. While I hate this outcome, I do believe we must cover this case:

Let's think about an http server. It might very well attend 20.000 request / minute per instance. This means a backend instance might very well attend 333 request / second. If a child container is created for each request, we should not traverse the array in every request.

Keeping this in mind, my final proposal is the following one: whenever `subscribe` is called, if `this.subscribers.lenght > 0 && this.subscribers.lenght % 1024 === 0` (we can extract 1024 to another constant), we traverse the subscribers array and, as before, if more than half the elements return `undefined` when dereferenced, then reallocate the array. This way flows like the one I just described keep the subscribers array healthy by traversing the array a reasonable amount of times. While `subscribe` is technically `O(n)`, I think we are avoiding a real performance degradation and we are solving the remaining memory leak in the subscriber array.

`WeakRef` constraint is unfortunate, but I agree with you: it's an extreme edge case. I cannot think about any real world scenario in which we are facing this.

leoptizmo:
With the changes to the `subscribe` method, running the test script with case (1) now yields the same results as inversify@7.0.0-alpha.1:

| `await job()` position  | Start | After loop GC | After container GC |
| ----------------------- | ----- | ------------- | ------------------ |
| Within loop (1)         | 4 MB  | 4 MB          | 4 MB               |
| Before loop GC (2)      | 4 MB  | 9 MB          | 4 MB               |
| Before container GC (3) | 4 MB  | 348 MB        | 4 MB               |
| N/A                     | 4 MB  | 348 MB        | 344 MB             |

Case (2) still crashes when run with `--max-old-space-size=200`.

notaphplover (approved):
LGTM!

PR #823, opened 2025-07-21
Update plan to reuse cached service nodes

notaphplover:
# Introduction

The current planner does generate plans whenever no existing plan is cached. Related services need to compute already computed subplans, adding complexity, affecting plan performance and incurring in a bigger memory usage. If we could make a smarter use of plan nodes, we could probably reuse them.

## What was currently preventing us from reusing service nodes?

Basically binding constraints. Given a service `s`, if a binding `b` associated to the service `s` has a constraint on top of its ancestors, we cannot reuse a plan built on top of `b` since there's no way to foresee wheter or not `b` is going to be used to provide the service `s` until we evaluate `b` constraint.

## What are doing to reuse service nodes

In the Chomsky grammar classification, G2 are denominated context free grammars. There's one interesting restriction for every production rule: the left-hand side must be a single nonterminal symbol. Basically, `S` can be produced regardless of any other symbol present in the current state.

The previous idea was the inspiration to say a `PlanServiceNode` is context free if an only if it can be planned regardless of the state of its ancestors. This PR updates the planner to efficiently mark service nodes as context free or not in order to populate the cache with additional plans that might be reused to compute subplans of non existing plans. Basically, we can track `BindingConstraints.getAncestor` calls to know, given a hierarchy of services, which ones are context free and which ones are not.

This approach has two main limitations:

1. As it might sound obvious, we can only reuse constext free service nodes.
2. We cannot reuse context free service nodes involving multiple tags for it would be expensive to access them later.

This approach has one important outcome: planning nodes no longer have parent for it would prevent us from reusing service nodes.

## What is the status of this PR
This PR is ready to be reviewed


packages/container/libraries/container/src/container/services/ServiceResolutionManager.ts
```diff
@@ -231,13 +231,19 @@ export class ServiceResolutionManager {
       getBindings: this.#getBindingsPlanParams,
       getBindingsChained: this.#getBindingsChainedPlanParams,
       getClassMetadata,
+      getPlan: this.#serviceReferenceManager.planResultCacheService.get.bind(
+        this.#serviceReferenceManager.planResultCacheService,
+      ),
```
notaphplover: We should compute this in the constructor so it's not computed on every plan call. This is probably adding some small overhead now

packages/container/libraries/container/src/container/services/ServiceResolutionManager.ts
```diff
@@ -231,13 +231,19 @@ export class ServiceResolutionManager {
...
       rootConstraints: this.#buildPlanParamsConstraints(
         serviceIdentifier,
         isMultiple,
         options,
       ),
       servicesBranch: [],
       setBinding: this.#setBindingParamsPlan,
+      setPlan: this.#serviceReferenceManager.planResultCacheService.set.bind(
+        this.#serviceReferenceManager.planResultCacheService,
+      ),
```
notaphplover: Same

PR #891, opened 2025-08-12
refactor(http-core): refactor HttpResponse to let Exception use

adrianmjim:
### Changed
- Refactored HttpResponse in order to use abtractions with Error
- Refactored InversifyHttpAdapter to use ErrorHttpAdapter
- Refactored success HttpReponses to use SuccessHttpResponse

### Added
- Added SuccessHttpResponse
- Added ErrorHttpResponse

### Motivation
We are changing approach of api errors to get a clear form

### DIFF
--- a/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
+++ b/packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
@@ -25,11 +25,12 @@ import { RouteParams } from '../models/RouteParams';
 import { RouterParams } from '../models/RouterParams';
 import { Pipe } from '../pipe/model/Pipe';
 import { PipeMetadata } from '../pipe/model/PipeMetadata';
-import { BadRequestHttpResponse } from '../responses/error/BadRequestHttpResponse';
+import { ErrorHttpResponse } from '../responses/error/ErrorHttpResponse';
 import { ForbiddenHttpResponse } from '../responses/error/ForbiddenHttpResponse';
 import { InternalServerErrorHttpResponse } from '../responses/error/InternalServerErrorHttpResponse';
 import { HttpResponse } from '../responses/HttpResponse';
 import { HttpStatusCode } from '../responses/HttpStatusCode';
+import { SuccessHttpResponse } from '../responses/success/SuccessHttpResponse';
 import { isPipe } from '../typeguard/isPipe';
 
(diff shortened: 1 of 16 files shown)


packages/http/libraries/core/src/http/responses/error/ErrorHttpResponse.ts
```diff
@@ -1,8 +1,33 @@
+import { Stream } from 'stream';
```
notaphplover:
```suggestion
import { Stream } from 'node:stream';
```

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -540,7 +525,7 @@ export abstract class InversifyHttpAdapter<
       undefined;
     let httpStatusCode: HttpStatusCode | undefined = statusCode;
 
-    if (HttpResponse.is(value)) {
+    if (SuccessHttpResponse.is(value) || ErrorHttpResponse.is(value)) {
```
notaphplover, line 528:
```suggestion
    if (HttpResponse.is(value)) {
```

I see the benefit of adding `SuccessHttpResponse.is` and `ErrorHttpResponse.is`, but keeping `HttpResponse.is` would be perfect here, wouldn't it?

notaphplover (commented):
I just added some comments. The code looks good :slightly_smiling_face:

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
adrianmjim: Yeah, that was my first approach but HttpResponse is not a class anymore. We could create a util function.

notaphplover: what is preventing us from being a class? I mean, if we let HttpResponse as it was and we extend a `HttpResponse` class instead of implementing a `HttpResponse` interface, we got the best of both worlds, right?

adrianmjim: The problem is at `ErrorHttpResponse`. Now, `ErrorHttpResponse` is extending `Error` so It can not extend `HttpResponse`

notaphplover: Ohh I see. Then we can leave it like this

notaphplover (approved):

PR #903, opened 2025-08-15
Remove Guard.getHttpResponse in favor of handling ErrorHttpResponse errors

notaphplover:
### Added
- Added `Guard` and `UseGuard` code examples.

### Changed
- Updated `InversifyHttpAdapter` to handle `ErrorHttpResponse` errors.

### Removed
- Removed `Guard.getHttpResponse`.

### DIFF
--- a/packages/docs/services/inversify-http-site/docs/api/_category_.json
+++ b/packages/docs/services/inversify-http-site/docs/api/_category_.json
@@ -1,4 +1,4 @@
 {
   "label": "API",
-  "position": 2
+  "position": 3
 }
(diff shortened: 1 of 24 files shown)


adrianmjim:
Hi,

I see that you are stating that the guard can return an `HttpResponse`, but the guard should only return a boolean that interrupts the flow or allows it to continue. With the new error system, it should allow exceptions to be thrown. These must be caught and handled.

notaphplover:
> Hi,
> 
> I see that you are stating that the guard can return an `HttpResponse`, but the guard should only return a boolean that interrupts the flow or allows it to continue. With the new error system, it should allow exceptions to be thrown. These must be caught and handled.

Ohh, I see. You're right! Then I'll go that way and update docs to request to throw an ErrorHttpResponse instead

PR #912, opened 2025-08-16
Update HttpResponse with isHttpResponse symbol

notaphplover:
### Added
- Added `isHttpResponse` function.

### Changed
- Updated `HttpResponse` with `isHttpResponse` symbol.
- Updated `InversifyHttpAdapter` to rely on `isHttpResponse`.

### DIFF
--- a/packages/http/libraries/core/src/http/responses/error/ErrorHttpResponse.ts
+++ b/packages/http/libraries/core/src/http/responses/error/ErrorHttpResponse.ts
@@ -1,16 +1,18 @@
 import { Stream } from 'node:stream';
 
-import { isHttpResponse } from '../calculations/isHttpResponse';
 import {
   HttpResponse,
-  isHttpResponse as isHttpResponseSymbol,
+  isHttpResponse as isHttpResponse,
 } from '../HttpResponse';
 import { HttpStatusCode } from '../HttpStatusCode';
 
-const httpErrorStatusCodeUpperBound: number = 600;
+const isErrorHttpResponse: unique symbol = Symbol.for(
+  '@inversifyjs/http-core/ErrorHttpResponse',
+);
 
 export class ErrorHttpResponse extends Error implements HttpResponse {
-  public readonly [isHttpResponseSymbol]: true;
+  public readonly [isHttpResponse]: true;
+  public readonly [isErrorHttpResponse]: true;
   public readonly body?: object | string | number | boolean | Stream;
 
   constructor(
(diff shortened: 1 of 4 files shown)


packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
```diff
@@ -329,11 +329,9 @@ export abstract class InversifyHttpAdapter<
       } catch (error: unknown) {
         this.#printError(error);
 
-        let response: HttpResponse = new InternalServerErrorHttpResponse();
-
-        if (ErrorHttpResponse.is(error)) {
-          response = error;
-        }
+        const response: HttpResponse = ErrorHttpResponse.is(error)
```
adrianmjim: Why do You prefer a ternary here?

packages/http/libraries/core/src/http/responses/error/ErrorHttpResponse.ts
```diff
@@ -1,14 +1,16 @@
...
 import { HttpStatusCode } from '../HttpStatusCode';
 
-const isErrorHttpResponseSymbol: unique symbol = Symbol.for(
-  '@inversifyjs/http-core/ErrorHttpResponse',
-);
+const httpErrorStatusCodeUpperBound: number = 600;
 
 export class ErrorHttpResponse extends Error implements HttpResponse {
-  public readonly [isErrorHttpResponseSymbol]: true;
+  public readonly [isHttpResponseSymbol]: true;
```
adrianmjim: I understand the approach being taken here. But perhaps we could keep the isErrorHttpResponseSymbol symbol to make the verification more explicit and efficient.

packages/http/libraries/core/src/http/responses/success/SuccessHttpResponse.ts
```diff
@@ -1,29 +1,27 @@
...
+} from '../HttpResponse';
 import { HttpStatusCode } from '../HttpStatusCode';
 
-const isSuccessHttpResponseSymbol: unique symbol = Symbol.for(
-  '@inversifyjs/http-core/SuccessHttpResponse',
-);
-
 export class SuccessHttpResponse implements HttpResponse {
-  public [isSuccessHttpResponseSymbol]: true;
+  public [isHttpResponseSymbol]: true;
```
adrianmjim: Same

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
notaphplover: To avoid initializing a new instance when it's not necessary. We could use an if else, but the ternary is just fine

packages/http/libraries/core/src/http/responses/error/ErrorHttpResponse.ts
notaphplover: I prefer this approach. We are likely to have a `RedirectionHttpResponse` for 300x status codes, this approach allows us to easily identify HttpResponses ~~, which I think it's interesting in the future of Pipes and Guards~~

packages/http/libraries/core/src/http/adapter/InversifyHttpAdapter.ts
adrianmjim: I don't like using ternary operators because of visibility issues when scaling code. But in this case, it's acceptable.

packages/http/libraries/core/src/http/responses/error/ErrorHttpResponse.ts
adrianmjim: I understand the approach, but by adding the symbol we would have the same result in a more explicit way and delegating at all times to the class type. In other words, it does not need to know any other data to perform the validation.

notaphplover: I think it's reasonable. Let's go that way :+1:

PR #925, opened 2025-08-18
Zod validation

notaphplover:
### Added
- Added `ZodValidationPipe`.
- Added `ValidateZodType` decorator.


packages/http/libraries/validation/zod/src/pipe/ZodValidationPipe.ts
```diff
@@ -0,0 +1,51 @@
+import { Pipe, PipeMetadata } from '@inversifyjs/http-core';
+import { BadRequestHttpResponse } from '@inversifyjs/http-core';
+import { getOwnReflectMetadata } from '@inversifyjs/reflect-metadata-utils';
+import { ZodSafeParseResult, ZodType } from 'zod';
+
+import { zodValidationMetadataReflectKey } from '../reflectMetadata/data/zodValidationMetadataReflectKey';
+
+export class ZodValidationPipe implements Pipe {
+  readonly #schemaList: ZodType[];
```
adrianmjim:
```suggestion
  readonly #typeList: ZodType[];
```

packages/http/libraries/validation/zod/package.json
```diff
@@ -0,0 +1,78 @@
...
+  "license": "MIT",
+  "main": "lib/cjs/index.js",
+  "module": "lib/esm/index.js",
+  "exports": {
+    ".": {
+      "import": "./lib/esm/index.js",
+      "require": "./lib/cjs/index.js"
+    }
+  },
+  "name": "@inversifyjs/zod-validation",
```
adrianmjim: I'm not sure if that's the most appropriate name. Ultimately, this package has to be used in the context of http or it won't work. Wouldn't it be better to name it in that context?

adrianmjim (commented):
Just a question

packages/http/libraries/validation/zod/package.json
notaphplover: Absolutely, you're right, what about `@inversifyjs/http-zod`?

packages/framework/libraries/validation/zod/src/pipe/ZodValidationPipe.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  #applySchemaList(input: unknown, typeList: ZodType[]): unknown {
+    let result: unknown = input;
+
+    for (const zodType of typeList) {
+      const parsedResult: ZodSafeParseResult<unknown> =
+        zodType.safeParse(result);
+
+      if (!parsedResult.success) {
+        throw new BadRequestHttpResponse(
```
adrianmjim: This is a ErrorHttpResponse. I think we have to remove this to use correctly http-core package

PR #934, opened 2025-08-21
feat(framework-core): add framework-core package and change references

adrianmjim:


packages/framework/libraries/core/src/index.ts
```diff
@@ -0,0 +1,5 @@
+import { Pipe } from './pipe/model/Pipe';
+import { PipeMetadata } from './pipe/model/PipeMetadata';
+import { isPipe } from './pipe/typeguard/isPipe';
+
+export { isPipe, Pipe, PipeMetadata };
```
notaphplover:
```suggestion
export { isPipe, type Pipe, type PipeMetadata };
```
Some bundlers find this extra piece of information really useful

packages/framework/libraries/core/README.md
```diff
@@ -0,0 +1,6 @@
+[![Test coverage](https://codecov.io/gh/inversify/monorepo/branch/main/graph/badge.svg?flag=%40inversifyjs%2Fhttp-core)](https://codecov.io/gh/inversify/monorepo/branch/main/graph/badge.svg?flag=%40inversifyjs%2Fhttp-core)
+[![npm version](https://img.shields.io/github/package-json/v/inversify/monorepo?filename=packages%2Fcontainer%2Flibraries%2Fhttp-core%2Fpackage.json&style=plastic)](https://www.npmjs.com/package/@inversifyjs/http-core)
+
+# @inversifyjs/http-core
+
+Inversify monorepo http core modules.
```
notaphplover: Update readme to target the right package

notaphplover (changes requested):
Nice, I see some missing changes:

1. Add changesets (major change).
2. What do we do with middlewares and Guards? They make sense outside http.
3. Update codecov config to register package tests.

.changeset/ripe-regions-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inversifyjs/framework-core": major
+---
+
+Create framework-core package
```
notaphplover: We should add one changeset per line to add in the changelog. I think we should add a changeset for every relevent export in the index file. Copilot can easily do it for you :)

notaphplover (approved):

PR #952, opened 2025-08-24
feat(framework-core): add interceptor support

adrianmjim:


packages/framework/libraries/core/src/interceptor/model/Interceptor.ts
notaphplover, line 1: Can we rename the folder to be `models`?

notaphplover (approved):
Lgtm

packages/framework/libraries/core/src/interceptor/model/Interceptor.ts
adrianmjim: Sure, my bad

notaphplover (approved):

PR #960, opened 2025-08-26
Add ExceptionFilter support 

adrianmjim:
### Added
- Add UseExceptionFilter decorator
- Add CatchException decorator

### Changed
- Move decorator and calculations to correct folders


packages/framework/libraries/core/src/exception-filter/calculations/exploreCatchExceptionMap.ts
```diff
@@ -0,0 +1,14 @@
+import { getOwnReflectMetadata } from '@inversifyjs/reflect-metadata-utils';
+import { Newable } from 'inversify';
+
+import { catchExceptionMetadataReflectKey } from '../../reflectMetadata/data/catchExceptionMetadataReflectKey';
+
+export function exploreCatchExceptionMap(): Map<
```
notaphplover: If we are going to rename stuff, can we use `get` instead of `explore`? I know nest uses `explore` in their codebase, but inversify container decorators uses `get` to provide metadata, like `getClassMetadata`

PR #969, opened 2025-08-31
refactor(http): move http packages to framework folder

adrianmjim:
### Changed
- Move http packages to framework folder

### Context
Recently, We have created a framework package and folder where we want to put every http package


notaphplover (changes requested):
vscode eslint config, codecov config and readme links should be updated accordingly.

I think it makes more sense to avoid the `libraries` parent folder: `packages/framework/http/libraries/*` instead of `packages/framework/libraries/http/libraries/*`. `http` packages might have non library packages.

notaphplover (approved):
lgtm

PR #1111, opened 2025-09-30
build: use consistent peer-deps requirements

BurningEnlightenment:
Unify the `reflect-metadata` version specification to `~0.2.2`. I chose to loosen the requirement as the dependents of `@inversifyjs/reflect-metadata-utils` consistently use the looser requirement.

Technically they don't need to match exactly; it'd be sufficient if the required version range does not get "looser" in downstream dependencies and provides sufficient overlap.

Which means the following examples would also be fine:
```
@inversifyjs/container requires reflect-metadata @ 0.2.2
└─ @inversifyjs/reflect-metadata-utils requires reflect-metadata @ ~0.2.2
```
```
@inversifyjs/container requires reflect-metadata @ 0.2.2
└─ @inversifyjs/reflect-metadata-utils requires reflect-metadata @ ~0.2.1
```
```
@inversifyjs/container requires reflect-metadata @ ~0.2.2
└─ @inversifyjs/reflect-metadata-utils requires reflect-metadata @ ~0.2.1
```

***

Btw. have you considered importing [`reflect-metadata/lite`](https://github.com/microsoft/reflect-metadata?tab=readme-ov-file#es-modules-in-nodejsbrowser-typescriptbabel-bundlers)? As far as I can tell all supported nodejs versions should be fine without the polyfills.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,9 @@
   "description": "A powerful and lightweight inversion of control container for JavaScript & Node.js apps powered by TypeScript",
   "devDependencies": {
     "@changesets/cli": "2.29.7",
-    "@commitlint/cli": "20.0.0",
+    "@commitlint/cli": "20.1.0",
     "@commitlint/config-conventional": "20.0.0",
-    "@commitlint/prompt-cli": "20.0.0",
+    "@commitlint/prompt-cli": "20.1.0",
     "@inversifyjs/foundation-changelog-generator": "workspace:*",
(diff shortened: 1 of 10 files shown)


notaphplover (commented):
You're absolutely right.

Can you please add a changeset to document the change? `pnpm changeset` should do it.

notaphplover (approved):

PR #1851, opened 2026-05-05
Update inversify build on top of tsgo

notaphplover:
### Changed
- Updated `inversify` build on top of `tsgo`.


notaphplover:
We will go for it when eslint compatibility is possible. The migration is not a real challenge.

PR #2107, opened 2026-09-01
feat(container): update jitless container option to be false by default

### LINKED ISSUE #2020: [Feature]: Update `jitless` container option to be false by default
### 🚀 Feature Proposal

`ContainerOptions.jitless` is `true` by default to avoid introducing breaking changes. Update it to `false` in `v9`.

### Motivation

`jit` related optimizations can speed up the resolution flow up to a 70%.

### Pitch

The monorepo takes care of `inversify`.

amandeavor:
Closes #2020. Updates DEFAULT_JITLESS to false by default for v9, adds unit tests in Container.spec.ts, updates docs, and adds changeset.

## Summary by CodeRabbit

* **Breaking Changes**
  * The container now uses JIT-optimized resolution by default.
  * Set `jitless: true` to use CSP-compatible resolution in environments that disallow `unsafe-eval`.

* **Documentation**
  * Updated English and Chinese API documentation to describe the new default behavior and configuration guidance.

### DIFF
--- a/.changeset/update-jitless-default-false.md
+++ b/.changeset/update-jitless-default-false.md
@@ -1,0 +1,6 @@
+---
+"@inversifyjs/container": major
+"inversify": major
+---
+
+- Updated `ContainerOptions.jitless` to default to `false`.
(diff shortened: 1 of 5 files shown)


notaphplover:
I'm afraid we don't want to go for this right now. V9 will be released in 2027, merging now would imply not releasing until then.

amandeavor:
Understood, that makes complete sense! I will close this for now so it doesn't block v8 release cycles with breaking changes, and we can revisit whenever the v9 milestone opens up. Thanks for clarifying the roadmap!

PR #2108, opened 2026-09-01
feat(http): provide Discriminated error support for ErrorFilter

### LINKED ISSUE #1025: [Feature]: [Http] Provide DiscriminatedError interface for better ErrorFilter support
## 🚀 Feature Proposal

`ErrorFilter` feature is implemented on top of a single not-so-true premise:

"Given `a` an instance of type `A`, `a instanceof A` is guaranteed to be true".

While this is true more often than not, there're some caveats. Phantom dependencies might lead to multiple `A` modules, multiple `A` constructors, meaning `a instanceof A` might be false in some unlucky scenarios.

It looks reasonable to provide an alternative way to deal with this issue.

### How do we deal with this problem in vanilla JavaScript?

This problem is not new, and neither is new a good enough solution to it: duck typing. This approach of asuming an object belongs to a given type if it has certain method / properties satisfying a set of conditions is my inspiration to solve the issue:

### Proposal: DiscriminatedError

If we can provide a hint to efficiently duck type an Error, we can avoid `instanceof` usage while accomplishing a performance friendly validation:

```ts
const errorDiscriminatorReflectKey: unique symbol = Symbol.for('@inversifyjs/validation-common/errorDiscriminator');

function Discriminated(value: string | symbol): ClassDecorator {
  return (target: Function): void {
    // Set property associated to target at errorDiscriminatorReflectKey
  }
}
```

This way, adapters can fetch target error discriminator metadata. If found, adapters can keep a map of discriminators to Errors. When an error is thrown, we can attempt to get the associated error getting it from its discriminator via accessing the error type metadata. In the worst case, we can always fall back to the `instanceof` approach if no Error is found this way.

#### Error hierarchies

Consider the following errors:

```ts
@Discriminated('foo')
class FooError extends Error {}

@Discriminated('foo-child')
class FooChildError extends FooError {}
```

Internally:
- `FooError` discriminator metadata is an array with one value: ['foo'].
- `FooChildError` discriminator metadata is an array with two values: ['foo-child', 'foo'].
- When a `FilterError` catches `FooError` errors, it's associated to the discriminator map to the key `foo`.
- When a `FilterError` catches `FooChildError` errors, it's associated to the discriminator map to the key `foo-child`.

When a `FooChildError` is thrown, the adapter reads it's constructor error discriminator metadata and finds ['foo-child', 'foo'].

1. The adapter first attemps to find filters for `foo-child`. If a `FilterError` handles `FooChildError`, it will be selected.
2. If not, the adapter attemps to find filters for `foo`. If a `FilterError` handles `FooError`, it will be selected.
3. If not, the adapter attemps to find Error filters with the current `instanceof` approach.

### Motivation

It solves an ErrorFilter related feature as described in the proposal.

### Example: Implementing a DiscriminatedError

```ts
const myAwesomeDiscriminatedErrorDiscriminator: string =
  'my-awesome-unique-discriminator-value';

@Discriminated(myAwesomeDiscriminatedErrorDiscriminator)
class MyAwesomeDiscrimatedError extends Error implements DiscriminatedError { }
```

#### Pros
- It can be implemented in a completelly transparent way. Developers implementing `ErrorFilter` classes for `FooError` don't need to care of whether or not the error is discriminated. Developers extending a `DiscriminatedError` don't need to be aware of its discriminated nature. Developers implementing a `DiscriminatedError` that extends a base class don't need to update the base class to be discriminated.
-  It's performant friendly: we can identify error filters in a few checks. We don't need to traverse the full type hierarchy.

#### Cons
- `ErrorFilter` is would be a feature whose implementation details matter. We could probably solve this enforcing every cached error must be discriminated, but the tradeof would be negative imo.

amandeavor:
Closes #1025. Implements @Discriminated decorator and DiscriminatedError interface, updates getErrorFilterForError and setErrorFilterToErrorFilterMap to support discriminator-based error lookup, adds unit tests, and adds changeset.

## Summary by CodeRabbit

* **New Features**
  * Added discriminator-based error filter matching using string or symbol identifiers.
  * Added the `@Discriminated` decorator for associating identifiers with error classes.
  * Error filters now support route-level and global discriminator lookup alongside type-based matching.
  * More specific error handlers take precedence during filter resolution.
  * Added public access to error discriminator metadata.
  * Discriminator metadata is evaluated per error class, preserving existing type-based behavior.


packages/framework/core/src/index.ts
```diff
@@ -16,6 +18,7 @@ export { ApplyMiddleware } from './middleware/decorators/ApplyMiddleware.js';
 export { MiddlewarePhase } from './middleware/models/MiddlewarePhase.js';
 export { isPipe } from './pipe/typeguard/isPipe.js';
 
+export type { DiscriminatedError } from './error/models/DiscriminatedError.js';
```
notaphplover: The interface is unused. The issue uses `DiscriminatedError` as part of a fictional codebase.

packages/framework/core/src/error/calculations/buildErrorDiscriminatorMetadata.ts
```diff
@@ -0,0 +1,11 @@
+export function buildErrorDiscriminatorMetadata(
+  discriminator: string | symbol,
+): (discriminators: (string | symbol)[]) => (string | symbol)[] {
+  return (discriminators: (string | symbol)[]): (string | symbol)[] => {
+    if (!discriminators.includes(discriminator)) {
+      discriminators.unshift(discriminator);
```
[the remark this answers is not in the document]
notaphplover: Agree

packages/framework/http/libraries/core/src/http/actions/setErrorFilterToErrorFilterMap.ts
```diff
@@ -17,17 +18,43 @@ export function setErrorFilterToErrorFilterMap(
...
-
-    if (existingErrorFilter === undefined) {
-      errorTypeToErrorFilterMap.set(errorType, errorFilter);
-    } else {
-      const errorTypeName: string =
-        errorType === null ? 'null (catch-all)' : errorType.name;
-      logger.warn(
-        `Error filter '${errorFilter.name}' was not registered for error type '${errorTypeName}' because an error filter is already registered for this error type.`,
-      );
+    const keysToRegister: (Newable<Error> | string | symbol | null)[] = [
```
notaphplover: Why is `keysToRegister` allowed to be `Newable<Error>` or `null`? It is populated with `discriminator` which is `string | symbol`

packages/framework/http/libraries/core/src/http/actions/setErrorFilterToErrorFilterMap.ts
```diff
@@ -17,17 +18,43 @@ export function setErrorFilterToErrorFilterMap(
...
+            keysToRegister.push(discriminator);
+          }
+        }
+      }
+    }
+
+    const errorTypeToFilterMap: Map<
+      unknown,
+      ErrorFilter | Newable<ErrorFilter>
+    > = errorTypeToErrorFilterMap;
```
notaphplover: No need to reassign this, right?

packages/framework/http/libraries/core/src/http/actions/setErrorFilterToErrorFilterMap.ts
```diff
@@ -17,17 +18,43 @@ export function setErrorFilterToErrorFilterMap(
...
+
+      if (existingErrorFilter === undefined) {
+        errorTypeToFilterMap.set(key, errorFilter);
+      } else if (key === errorType) {
+        const errorTypeName: string =
+          errorType === null ? 'null (catch-all)' : errorType.name;
+
+        logger.warn(
+          `Error filter '${errorFilter.name}' was not registered for error type '${errorTypeName}' because an error filter is already registered for this error type.`,
+        );
```
notaphplover: We shouldn't warn, collisions should be expected and documented instead

packages/framework/http/libraries/core/src/http/calculations/getErrorFilterForError.ts
```diff
@@ -1,10 +1,30 @@
...
 
-function* getErrorBaseTypeChain(
+function* getErrorMatchCandidates(
   error: unknown,
-): Generator<Newable<Error> | null> {
+): Generator<Newable<Error> | string | symbol | null> {
+  if (typeof error === 'object' && error !== null) {
+    const errorConstructor: NewableFunction | undefined = (
+      error as { constructor?: NewableFunction }
+    ).constructor;
```
notaphplover: `getErrorDiscriminatorMetadata` admits `Function`, we can simplify this:

```ts
const errorConstructor: Function = error.constructor;
```

No need to check for undefined `errorConstructor`, right? If `error` is a non null object, then it has a `constructor` which is the constructor of the instance.

packages/framework/http/libraries/core/src/http/calculations/getErrorFilterForError.ts
```diff
@@ -1,10 +1,30 @@
-import { type ErrorFilter } from '@inversifyjs/framework-core';
+import {
+  type ErrorFilter,
+  getErrorDiscriminatorMetadata,
+} from '@inversifyjs/framework-core';
 import { getBaseType } from '@inversifyjs/prototype-utils';
 import { type Container, type Newable } from 'inversify';
 
-function* getErrorBaseTypeChain(
+function* getErrorMatchCandidates(
```
notaphplover: what is the purpose of using a generator?

packages/framework/http/libraries/core/src/http/calculations/getErrorFilterForError.ts
```diff
@@ -27,10 +47,12 @@ export async function getErrorFilterForError<TRequest, TResponse, TResult>(
     ErrorFilter | Newable<ErrorFilter>
   >[],
 ): Promise<ErrorFilter<unknown, TRequest, TResponse, TResult> | undefined> {
-  for (const errorType of getErrorBaseTypeChain(error)) {
+  for (const candidate of getErrorMatchCandidates(error)) {
```
notaphplover: I think we're breaking non discriminated errors this way. Old error filters should continue working. If we no longer traverse the error type chain, how are we going to find error filters for the base type in the event no discriminated error filters are found?

notaphplover (changes requested):
I have some doubts regarding the feature implementation. Let's solve them before merging.

packages/framework/http/libraries/core/src/http/actions/setErrorFilterToErrorFilterMap.ts
notaphplover: Ohh, the initial value. Understood

packages/framework/http/libraries/core/src/http/calculations/getErrorFilterForError.ts
notaphplover: Not really harmful, we can keep it

packages/framework/http/libraries/core/src/http/calculations/getErrorFilterForError.ts
notaphplover: Nope, I'm absolutely wrong on this

packages/framework/http/libraries/core/src/http/actions/setErrorFilterToErrorFilterMap.ts
notaphplover: Ohh, we are tricking the compiler here, we should not do this.

`errorTypeToErrorFilterMap` is no longer an `error type to error filter map`  and it no longer contains only `Newable<Error> | null`. Why are we mixing discriminators with error types? Wouldn't it be simpler and cleaner to have another map for error discriminators to error filters map?

packages/framework/http/libraries/core/src/http/calculations/getErrorFilterForError.ts
```diff
@@ -27,10 +47,12 @@ export async function getErrorFilterForError<TRequest, TResponse, TResult>(
...
     for (const errorToFilterMap of errorToFilterMapList) {
       const errorFilterOrType: ErrorFilter | Newable<ErrorFilter> | undefined =
-        errorToFilterMap.get(errorType);
+        (
+          errorToFilterMap as Map<unknown, ErrorFilter | Newable<ErrorFilter>>
```
notaphplover: This trick is the outcome of previously tricking the compiler, er should avoid this

amandeavor:
Thanks for the review. I pushed the fixes: discriminator filters now use a separate map, the unused interface is gone, inherited metadata is copied instead of mutated, and normal error-type inheritance still falls back through the existing chain. I also updated the tests and changeset. Both affected packages build cleanly, all 366 package unit tests pass, and the pre-commit changed-file tests and type checks pass too.

notaphplover (approved):

